PR #16, opened 2013-02-17
[Fixed JENKINS-16847] Added explicit support for copy-artifact steps being part of MultiJob pi...

trojanbug:
...pelines, to detect other builds finished as part of the same multijob build as "upstream".
This allows using "Upstream project that triggered this build" for artifact selection and making sure it copies the artifact from the right build (before the only solution was "last successful build", for longer pipelines unacceptable as several builds could happen "in between".

A recursive algorithm will descend also to parent builds being multijobs themselves.
It is yet to be tested if this also will work "up" the chain (https://github.com/jenkinsci/copyartifact-plugin/pull/15 suggests not, while mu understanding of of getCauses() contract suggest it will - to be tested).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,6 +26,12 @@
         <artifactId>maven-plugin</artifactId>
         <optional>true</optional>
       </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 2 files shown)


MarkEWaite:
Multijob plugin is deprecated.  Closing

PR #17, opened 2013-02-25
Support standard Run parameters to select build instead of only custom ParameterizedBuildSelector

petehayes:
I had a use case where I wanted to use a standard run parameter as part of my job as well as the copy artifacts plugin.  The copy artifacts plugin requires that you use the custom parameterized build selector so I would end up defining the parameter twice.  This pull request adds support in the Copy artifacts plugin to look for standard Jenkins Run parameter and uses that if found, otherwise it continues with the current behavior.


ikedam:
As described in [JENKINS-21519](https://issues.jenkins-ci.org/browse/JENKINS-21519), you can already use run parameters with specific build selectors and ${PARAMETER_NAME_NUMBER}.
No need to extend copyartifact.

PR #32, opened 2014-02-20
RFC: Record copied artifacts in build action

daniel-beck:
Which artifacts are copied in a certain build is as important as knowing the exact SCM revisions it is based on. So it should be displayed as prominently.
- Show list of source projects and builds on build page
- Add new sidepanel entry with detail page listing copied files

Screenshots below.

---

This is a request for comments on this implementation and shouldn't be merged yet. Is this complete? I'm not using matrix or Maven projects, so I'm really interested in information on those.

(Work in progress: weird difference needed to cover both flattened and unflattened copying; unknown whether `Run.XSTREAM2` is the right place to define aliases; insane approach to getting the current build's `sidepanel.jelly`; no full localizability; no tests; not making sure anchor names are valid)

---

screen shot 2014-02-20 at 01 07 20

---

screen shot 2014-02-20 at 01 08 01

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -519,6 +519,11 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
             return isOptional();  // Fail build unless copy is optional
         }
 
+        CopiedArtifactsRunAction a = dst.getAction(CopiedArtifactsRunAction.class);
+        if (a == null) {
+            a = new CopiedArtifactsRunAction();
+            dst.addAction(a);
+        }
         copier.initialize(src, dst, srcDir, baseTargetDir);
         try {
             int cnt;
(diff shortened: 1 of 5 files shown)


daniel-beck:
Rebased onto current master. (Original was a236cac88ea8e8d597c6c931b002606273cb2ef4)

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -527,8 +532,11 @@ private boolean perform(Run src, Run<?,?> dst, String expandedFilter, String exp
             else {
                 targetDir.mkdirs();  // Create target if needed
                 FilePath[] list = srcDir.list(expandedFilter, expandedExcludes, false);
-                for (FilePath file : list)
+                for (FilePath file : list) {
                     copier.copyOne(file, new FilePath(targetDir, file.getName()), isFingerprintArtifacts());
+                    String relativePath = file.getRemote().substring(srcDir.getRemote().length() + 1);
+                    a.recordSourceFile(src, relativePath);
```
jglick: You may wish to `dst.save()` at this point.

PR #41, opened 2014-06-19
DO NOT MERGE: [JENKINS-23475] tests to demonstrate JENKINS-23475.

ikedam:
These are tests to reproduce [JENKINS-23475](https://issues.jenkins-ci.org/browse/JENKINS-23475), permission check can be bypassed with WebAPI (and also with CLI).
Specifying project names without variables (that is, statically), permission checks are performed at configuration time, but they are not performed for WebAPI and CLI.

Test will fail for this pull request.


ikedam:
Fixed in SECURITY-988, copyartifact-1.44
https://github.com/jenkinsci/copyartifact-plugin/blob/master/CHANGELOG.adoc#144

PR #53, opened 2014-12-17
Add new "last build" selector

hschaa:
Allows to select the last build ignoring the build status completely.
This allows to copy artifacts also from failed jobs.


hschaa:
Updated pull request without ordinal=100.

ikedam:
Released copyartifact-1.33.
It will be available in a day.

PR #55, opened 2015-01-13
[FIXED JENKINS-24887] CopyArtifact to implement workflow SimpleBuildStep

tfennelly:
Contains first set of changes aimed at making the CopyArtifact build step usable within a workflow. Supersedes #43 and implements [JENKINS-24887](https://issues.jenkins-ci.org/browse/JENKINS-24887).

* Existing tests all pass.
* There are a number of __TODO__ question marks that I think need to be resolved.


src/main/java/hudson/plugins/copyartifact/Copier.java
```diff
@@ -35,7 +35,7 @@
      * @param baseTargetDir Base target dir for upcoming file copy (the copy-artifact
      *   build step may later specify a deeper target dir)
      */
-    public abstract void init(Run src, AbstractBuild<?,?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException;
+    public abstract void init(Run src, Run<?,?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException;
```
jglick: Possible backward-compatibility issue. I am not sure if there is really an expectation of an implementation of this extension point in another plugin, but if so, you need to do some gymnastics with `Util.isOverridden` to make it compatible to both call and override the original overload.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -124,7 +126,14 @@ public CopyArtifact(String projectName, String parameters, BuildSelector selecto
 
         this.project = projectName;
         this.parameters = Util.fixEmptyAndTrim(parameters);
-        this.selector = selector;
+
+        if (selector == null) {
+            // TODO: specify the BuildSelector instance in the workflow script via CoreStep?
+            this.selector = new StatusBuildSelector(true);
```
jglick: Use `@DataBoundSetter` to define a default value for a property. [Tips](https://github.com/jenkinsci/workflow-plugin/blob/master/scm-step/README.md#constructor-vs-setters)

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -260,14 +269,25 @@ public boolean isFingerprintArtifacts() {
     public boolean perform(AbstractBuild<?,?> build, Launcher launcher, BuildListener listener)
```
jglick: This overload can simply be deleted when implementing `SimpleBuildStep`.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -372,14 +397,17 @@ private boolean canReadFrom(Job<?, ?> job, AbstractBuild<?, ?> build) {
...
-    private ItemGroup getItemGroup(AbstractBuild<?, ?> build) {
-        ItemGroup group = build.getProject().getRootProject().getParent();
-        return group;
-
+    private ItemGroup getItemGroup(Run<?, ?> build) {
+        if (build instanceof AbstractBuild) {
+            return ((AbstractBuild) build).getProject().getRootProject().getParent();
+        } else {
+            // TODO: hmmm ???
+            return build.getParent().getParent();
```
jglick: This is fine. There is no equivalent of `getRootProject` for non-`AbstractProject`s, and none is needed. (The original code here is just the usual hacks to make things work with `matrix-project`.)

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -282,10 +302,15 @@ public boolean perform(AbstractBuild<?,?> build, Launcher launcher, BuildListene
...
                 return isOptional();  // Fail build unless copy is optional
             }
-            FilePath targetDir = build.getWorkspace(), baseTargetDir = targetDir;
+            FilePath targetDir = workspace, baseTargetDir = targetDir;
             if (targetDir == null || !targetDir.exists()) {
-                console.println(Messages.CopyArtifact_MissingWorkspace()); // (see JENKINS-3330)
-                return isOptional();  // Fail build unless copy is optional
+                if (targetDir != null) {
+                    // TODO: What do we do if the workspace dir does not exist, as in the case of workflow (it seems)? Forcing create for now.
+                    targetDir.mkdirs();
```
jglick: When using `SimpleBuildStep`, it is guaranteed that `workspace` is not null and exists.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -337,9 +362,9 @@ public boolean perform(AbstractBuild<?,?> build, Launcher launcher, BuildListene
         }
     }
 
-    private boolean canReadFrom(Job<?, ?> job, AbstractBuild<?, ?> build) {
-        if ((job instanceof AbstractProject) && CopyArtifactPermissionProperty.canCopyArtifact(
-                build.getProject().getRootProject(),
+    private boolean canReadFrom(Job<?, ?> job, Run<?, ?> build) {
+        if ((job instanceof AbstractProject) && (build instanceof AbstractBuild) && CopyArtifactPermissionProperty.canCopyArtifact(
+                ((AbstractBuild)build).getProject().getRootProject(),
                 ((AbstractProject<?,?>)job).getRootProject()
```
jglick: Still want to do the `canCopyArtifact` check on non-`AbstractBuild`s, just do not need to call `getRootProject`.

src/main/java/hudson/plugins/copyartifact/FingerprintingCopyMethod.java
```diff
@@ -43,9 +43,9 @@
     private final Map<String,String> fingerprints = new HashMap<String, String>();
 
     @Override
-    public void init(Run src, AbstractBuild<?, ?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException {
+    public void init(Run src, Run<?, ?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException {
         this.src = src instanceof AbstractBuild ? (AbstractBuild)src : null;
-        this.dst = dst;
+        this.dst = dst instanceof AbstractBuild ? (AbstractBuild)dst : null; // TODO: hmmm ?
```
jglick: Need to switch the fields here to be `Run`s, and call `Fingerprint.addFor`.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
tfennelly: Thanks, will do.

Also wondering how would users specify different `BuildSelector` types  from within the workflow script?

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
tfennelly: There's a boolean return on the original? Maybe you mean something else.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
tfennelly: I added this because I was not getting a workspace in the initial tests i.e. the workspace dir did not exist and the step was exiting early.

src/main/java/hudson/plugins/copyartifact/FingerprintingCopyMethod.java
tfennelly: Sure. Just trying to minimize changes as instructed.

tfennelly:
@jglick Thanks for the comments, Jesse. Will make the changes.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: Use the `ScriptGenerator` to see. Something like

    selector: [$class: 'StatusBuildSelector', something: true]

src/main/java/hudson/plugins/copyartifact/Copier.java
jglick: https://github.com/jenkinsci/git-plugin/blob/66f632846c21886eef1180daff246d9e0ab973ca/src/main/java/hudson/plugins/git/extensions/GitSCMExtension.java#L65-83 for example.

jglick: https://github.com/jenkinsci/jenkins/commit/cb6e568815976a2309bff9ebf3e4f7e0a070e769#diff-f53a4bf03f87b35d03b5c29476b723edR42 is the one case where I can remember actually testing this properly.

jglick: https://github.com/jenkinsci/jenkins/commit/5b6522808090ae675826a6033382f2361f7f270c was the original change.

pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.509</version>
+        <version>1.577</version>
```
jglick: Suggest 1.580.1.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: Returning anything other than `true` was already deprecated.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: Right, so you can just `targetDir.mkdirs()` unconditionally and quietly.

src/main/java/hudson/plugins/copyartifact/FingerprintingCopyMethod.java
jglick: Changing this class to fully support `Run` is actually needed, however.

tfennelly:
@jglick made the updates we discussed. Maybe you could have a quick scan again.

In particular: 

• In e00e48c you'll see I had to explicitly include a dep on jenkins remoting.
* The `Util.isOverridden` shenanigans is in 0ef1cfa.

tfennelly:
`TriggeredBuildSelectorTest` failing on CI (not failing locally).

Might rollback commits and see which is causing the CI failure.

tfennelly:
Ok, CI passes after adding the commits one at a time. Some random issue.

pom.xml
```diff
@@ -22,10 +22,24 @@
...
     <dependencies>
+      <!--
+        TODO: ??
+        After updating parent pom version dep to 1.580.1, we needed to explicitly include
+        the remoting jar before maven-plugin. If not, we get a 'java.lang.SecurityException: class
+        "org.jenkinsci.remoting.CallableDecorator"'s signer information does not match signer
+        information of other classes in the same package' exception. Looks to be related to the
+        addition of org.jenkins-ci:SECURITY-144-compat:jar:1.0, which the maven plugin depends on
+        and has classes in the org.jenkinsci.remoting package.
+      -->
```
jglick: This is [JENKINS-25625](https://issues.jenkins-ci.org/browse/JENKINS-25625) (CC @kohsuke).

src/main/java/hudson/plugins/copyartifact/Copier.java
```diff
@@ -34,8 +39,24 @@
...
      * @param baseTargetDir Base target dir for upcoming file copy (the copy-artifact
      *   build step may later specify a deeper target dir)
+     *
+     * @since TODO ?
      */
-    public abstract void init(Run src, AbstractBuild<?,?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException;
+    public void init(Run src, Run<?,?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException {
+        if (dst instanceof AbstractBuild && Util.isOverridden(Copier.class, getClass(), "init", Run.class, AbstractBuild.class, FilePath.class, FilePath.class)) {
+            init(src, (AbstractBuild) dst, srcDir, baseTargetDir);
+        }
```
jglick: else throw `AbstractMethodError`

src/main/java/hudson/plugins/copyartifact/Copier.java
```diff
@@ -34,8 +39,24 @@
...
+    }
+
+    /**
+     * @deprecated Please use {@link #init(hudson.model.Run, hudson.model.Run, hudson.FilePath, hudson.FilePath)}
+     */
+    @Deprecated
+    public void init(Run src, AbstractBuild<?,?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException {
+        if (Util.isOverridden(Copier.class, getClass(), "init", Run.class, Run.class, FilePath.class, FilePath.class)) {
+            init(src, (Run) dst, srcDir, baseTargetDir);
+        }
```
jglick: else throw `AbstractMethodError`

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -63,35 +62,40 @@
...
     private String parameters;
-    private final String filter, target;
-    private final String excludes;
+    private String filter, target;
+    private String excludes;
     private /*almost final*/ BuildSelector selector;
     @Deprecated private transient Boolean stable;
-    private final Boolean flatten, optional;
-    private final boolean doNotFingerprintArtifacts;
+    private Boolean flatten, optional;
```
jglick: I see no reason for these to be `Boolean`. I think they can be changed to `boolean`. The only behavioral change should be that the XStream form for the default configuration will then include

```
<flatten>false</flatten>
<optional>false</optional>
```

but that seems fine.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -105,9 +109,22 @@ public CopyArtifact(String projectName, String parameters, BuildSelector selecto
...
+        setFlatten(flatten);
+        setOptional(optional);
+        setFingerprintArtifacts(fingerprintArtifacts);
+    }
+
+    @DataBoundConstructor
+    public CopyArtifact(String projectName) {
         // check the permissions only if we can
         StaplerRequest req = Stapler.getCurrentRequest();
         if (req!=null) {
```
jglick: BTW this block is legacy code, predating `QueueItemAuthenticator`; I thought @ikedam had already cleaned the permission system in this plugin up, but perhaps there is still some junk left behind. At any rate, it should be a no-op for Workflow since this constructor is called during the build, when there is no active web request. (It can _also_ be called by the snippet generator, when there is, but then there will be no `AbstractProject` in the context—only a `[Workflow]Job`.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -257,36 +316,46 @@ public boolean isFingerprintArtifacts() {
...
+            env.overrideAll(((AbstractBuild)build).getBuildVariables()); // Add in matrix axes..
+        }
+
         PrintStream console = listener.getLogger();
         String expandedProject = project, expandedFilter = filter;
         String expandedExcludes = getExcludes();
         try {
-            EnvVars env = build.getEnvironment(listener);
-            env.overrideAll(build.getBuildVariables()); // Add in matrix axes..
             expandedProject = env.expand(project);
```
jglick: This call to `expand` is at best useless for a workflow, which should already have done any variable expansion it wanted before calling the constructor.

I think it would suffice to set `env` to be an empty `new EnvVars()` if not dealing with an `AbstractBuild`, so that calls to `expand` here and elsewhere just become the identity function. (Just double-check that this is the only thing `Selector.getBuild` uses `env` for.)

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -371,15 +451,25 @@ private boolean canReadFrom(Job<?, ?> job, AbstractBuild<?, ?> build) {
         return b;
     }
 
-    // retrieve the "folder" (jenkins root if no folder used) for this build
-    private ItemGroup getItemGroup(AbstractBuild<?, ?> build) {
-        ItemGroup group = build.getProject().getRootProject().getParent();
-        return group;
+    private Job<?, ?> getRootJob(Job<?, ?> job) {
```
jglick: More intuitively named `getRootProject`, and can be `static`.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -371,15 +451,25 @@ private boolean canReadFrom(Job<?, ?> job, AbstractBuild<?, ?> build) {
...
+            return job;
+        }
+    }
 
+    // retrieve the "folder" (jenkins root if no folder used) for this build
+    private ItemGroup getItemGroup(Run<?, ?> build) {
+        if (build instanceof AbstractBuild) {
+            return ((AbstractBuild) build).getProject().getRootProject().getParent();
+        } else {
+            return build.getParent().getParent();
```
jglick: Can be simplified with the method you just added:

    return getRootProject(build.getParent()).getParent();

src/main/java/hudson/plugins/copyartifact/CopyArtifactException.java
```diff
@@ -0,0 +1,40 @@
...
+ * THE SOFTWARE.
+ */
+package hudson.plugins.copyartifact;
+
+import java.io.IOException;
+
+/**
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public class CopyArtifactException extends IOException {
```
jglick: Useless. Just use `AbortException`.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -257,36 +316,46 @@ public boolean isFingerprintArtifacts() {
...
                 // If projectName is parameterized, need to do permission check on source project.
                 && !canReadFrom(job, build)) {
                 job = null; // Disallow access
             }
             if (job == null) {
-                console.println(Messages.CopyArtifact_MissingProject(expandedProject));
-                return false;
+                String message = Messages.CopyArtifact_MissingProject(expandedProject);
+                console.println(message);
+                throw new CopyArtifactException(message);
```
jglick: If you are throwing an `AbortException`, it is redundant to print the message. Just

```
if (job == null) {
    throw new AbortException(Messages.CopyArtifact_MissingProject(expandedProject));
}
```

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -257,36 +316,46 @@ public boolean isFingerprintArtifacts() {
...
+                console.println(message);
+                if (isOptional()) {
+                    // just return without an error
+                    return;
+                } else {
+                    // Fail build if copy is not optional
+                    throw new CopyArtifactException(message);
+                }
             }
+            FilePath targetDir = workspace, baseTargetDir = targetDir;
```
jglick: Presumably only one of these is mutated, so would suffice to rename it to `workspace` and remove the redundant local var.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -315,7 +384,9 @@ public boolean perform(AbstractBuild<?,?> build, Launcher launcher, BuildListene
                     Run<?,?> r = it.next();
                     ok |= perform(r, build, expandedFilter, expandedExcludes, targetDir, baseTargetDir, copier, console);
                 }
-                return ok;
+                if (!ok) {
+                    throw new CopyArtifactException(Messages.CopyArtifact_FailedToCopy(expandedProject, expandedFilter));
```
jglick: Can be inlined into the loop. I see no reason to continue copying if one iteration fails.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -371,15 +451,25 @@ private boolean canReadFrom(Job<?, ?> job, AbstractBuild<?, ?> build) {
...
+        if (build instanceof AbstractBuild) {
+            return ((AbstractBuild) build).getProject().getRootProject().getParent();
+        } else {
+            return build.getParent().getParent();
+        }
     }
 
 
-    private boolean perform(Run src, AbstractBuild<?,?> dst, String expandedFilter, String expandedExcludes, FilePath targetDir,
+    private boolean perform(Run src, Run<?,?> dst, String expandedFilter, String expandedExcludes, FilePath targetDir,
```
jglick: Just make this return `void` and throw `AbortException` as needed.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -324,24 +395,33 @@ public boolean perform(AbstractBuild<?,?> build, Launcher launcher, BuildListene
...
+            throw cae;
+        }
         catch (IOException ex) {
             Util.displayIOException(ex, listener);
-            ex.printStackTrace(listener.error(
-                    Messages.CopyArtifact_FailedToCopy(expandedProject, expandedFilter)));
-            return false;
+            String message = Messages.CopyArtifact_FailedToCopy(expandedProject, expandedFilter);
+            ex.printStackTrace(listener.error(message));
+            throw new CopyArtifactException(message);
```
jglick: This whole `catch` block could simply be deleted, I think: let the `IOException` be thrown up to the caller and handled there.

src/test/java/hudson/plugins/copyartifact/CopierTest.java
```diff
@@ -0,0 +1,120 @@
...
+        // not forward the call to C3/C4.init(Run, AbstractBuild) because the supplied runDst
+        // is not an AbstractBuild instance.
+        testRunDst(new C3(), 0);
+        testRunDst(new C4(), 0);
+    }
+
+    public void testRunDst(C c, int expectedCnt) throws IOException, InterruptedException {
+        Run runSrc = Mockito.mock(Run.class);
+        Run runDst = Mockito.mock(Run.class);
+        c.init(runSrc, runDst, null, null);
```
jglick: I do not think this is testing what you think, since it is always calling the new overload. And there are different behaviors when the runtime type of the argument is actually `AbstractBuild`, which you would need to `mock(AbstractBuild.class)` to demonstrate.

src/test/java/hudson/plugins/copyartifact/CopierTest.java
```diff
@@ -0,0 +1,120 @@
...
+    private class C3 extends C {
+        // Override init that uses the deprecated AbstractBuild dst
+        @Override
+        public void init(Run src, AbstractBuild<?, ?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException {
+            callCnt++;
+        }
+    }
+
+    private class C4 extends C {
+        // Override init that uses the deprecated AbstractBuild dst + call super
```
jglick: Should also test overriding the new overload and calling `super`. There is no reason to do so, of course (the method is logically abstract) but someone might.

By contrast, the use case you are testing _here_ is pointless for the purposes of ensuring that this change brings no regression: the old method _was_ `abstract`, so no existing implementations could have been calling `super`.

src/test/java/hudson/plugins/copyartifact/testutils/CopyArtifactUtil.java
```diff
@@ -0,0 +1,60 @@
...
+        return createCopyArtifact(projectName, parameters, selector, filter, null, target, flatten, optional, false);
+    }
+
+    public static CopyArtifact createCopyArtifact(String projectName, String parameters, BuildSelector selector, String filter, String target,
+            boolean flatten, boolean optional, boolean fingerprintArtifacts) {
+        return createCopyArtifact(projectName, parameters, selector, filter, null, target, flatten, optional, fingerprintArtifacts);
+    }
+
+    public static CopyArtifact createCopyArtifact(String projectName, String parameters, BuildSelector selector, String filter, String excludes, String target,
+                        boolean flatten, boolean optional, boolean fingerprintArtifacts) {
```
jglick: OK for purposes of creating a minimal diff, though in the long run I would expect this class to be deleted (inlined), with calls to default field values deleted.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -324,24 +395,33 @@ public boolean perform(AbstractBuild<?,?> build, Launcher launcher, BuildListene
...
-        if ((job instanceof AbstractProject) && CopyArtifactPermissionProperty.canCopyArtifact(
-                build.getProject().getRootProject(),
-                ((AbstractProject<?,?>)job).getRootProject()
-        )) {
+    private boolean canReadFrom(Job<?, ?> job, Run<?, ?> build) {
+        Job<?, ?> fromJob = job;
+        Job<?, ?> toJob = build.getParent();
+
+        if ((job instanceof AbstractProject) && (build instanceof AbstractBuild) &&
+                CopyArtifactPermissionProperty.canCopyArtifact(getRootJob(toJob), getRootJob(fromJob))) {
```
jglick: You have already fixed `canCopyArtifact` to accept `Job` so there is no need to check the types of the arguments.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -257,36 +316,46 @@ public boolean isFingerprintArtifacts() {
...
-                console.println(Messages.CopyArtifact_MissingWorkspace()); // (see JENKINS-3330)
-                return isOptional();  // Fail build unless copy is optional
+                String message = Messages.CopyArtifact_MissingBuild(expandedProject);
+                console.println(message);
+                if (isOptional()) {
+                    // just return without an error
+                    return;
+                } else {
+                    // Fail build if copy is not optional
+                    throw new CopyArtifactException(message);
```
jglick: Of course this should be

```
if (isOptional()) {
    console.println(message);
    return;
} else {
    throw new AbortException(message);
}
```

jglick:
> Might rollback commits and see which is causing the CI failure.

No need to be that extreme. Just push any additional commit to see if something is a random failure (which is not at all unusual for Jenkins code, I am afraid).

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
tfennelly: I did not make these fields `Boolean`. This is how they were in the original code. I don't see any reason for them not being `boolean` either but would prefer to check with @ikedam before making any changes. Better still, perhaps do it in a separate PR since it's not directly related to what we're trying to do here.

jglick: Related to the `@DataBoundSetter` changes, but true that it is orthogonal to this PR.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
tfennelly: That function is doing things a little strangely and I did not change the basic flow and would prefer not to without first consulting with @ikedam.

I will create a JIRA for this and we can look into it in a separate set of changes.

tfennelly:
@ikedam @jglick I think/hope this is ready to merge upstream.

@ikedam would you mind looking through these changes? Let me know if it makes it easier for you if I rebase/squash them all into a single commit.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: I just figured that since in the public overload of `perform` we are changing the signature to use a meaningful `AbortException` rather than the deprecated `false`, it makes sense to change it here to match.

jglick:
Presumably you no longer consider this a WiP.

src/main/java/hudson/plugins/copyartifact/Copier.java
```diff
@@ -34,8 +39,34 @@
      * @param srcDir Source for upcoming file copy
      * @param baseTargetDir Base target dir for upcoming file copy (the copy-artifact
      *   build step may later specify a deeper target dir)
+     *
+     * @since TODO ?
      */
-    public abstract void init(Run src, AbstractBuild<?,?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException;
+    public void init(Run src, Run<?,?> dst, FilePath srcDir, FilePath baseTargetDir) throws IOException, InterruptedException {
```
jglick: Avoid rawtypes: `Run<?,?> src`

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -123,13 +141,55 @@ public CopyArtifact(String projectName, String parameters, BuildSelector selecto
...
+    }
+
+    @DataBoundSetter
+    public void setExcludes(String excludes) {
+        this.excludes = Util.fixNull(excludes).trim();
+    }
+
+    @DataBoundSetter
+    public void setSelector(BuildSelector selector) {
+        this.selector = (selector != null ? selector : DEFAULT_BUILD_SELECTOR);
```
jglick: I guess your backout/rebase lost my earlier comment. In general, any push to a PR which invalidates a portion of the diff will automatically hide any comments made against that earlier diff.

Simplify this setter to assume a `@Nonnull` argument (the one call from the deprecated constructor should then check for null of course) and instead use `DEFAULT_BUILD_SELECTOR` as the field initializer. This is potentially important for [JENKINS-25779](https://issues.jenkins-ci.org/browse/JENKINS-25779).

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -257,91 +317,103 @@ public boolean isFingerprintArtifacts() {
...
+                // just return without an error
+                console.println(message);
+                return;
+            } else {
+                // Fail build if copy is not optional
+                throw new AbortException(message);
             }
-            FilePath targetDir = build.getWorkspace(), baseTargetDir = targetDir;
-            if (targetDir == null || !targetDir.exists()) {
-                console.println(Messages.CopyArtifact_MissingWorkspace()); // (see JENKINS-3330)
```
jglick: Assuming this was the only usage, you can now delete the key from `Messages.properties`.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -257,91 +317,103 @@ public boolean isFingerprintArtifacts() {
...
-    private boolean canReadFrom(Job<?, ?> job, AbstractBuild<?, ?> build) {
-        if ((job instanceof AbstractProject) && CopyArtifactPermissionProperty.canCopyArtifact(
-                build.getProject().getRootProject(),
-                ((AbstractProject<?,?>)job).getRootProject()
-        )) {
+    private boolean canReadFrom(Job<?, ?> job, Run<?, ?> build) {
+        Job<?, ?> fromJob = job;
+        Job<?, ?> toJob = build.getParent();
+
+        if ((job instanceof AbstractProject) && (build instanceof AbstractBuild) &&
```
jglick: Why the `instanceof` checks?

src/test/java/hudson/plugins/copyartifact/CopierTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+        // C3 only implements the legacy AbstractBuild init method (and not the Run
+        // version). Copier.init(Run, Run) will get called, but should throw an AbstractMethodError
+        try {
+            testRunDst(new C3(), 0);
+            Assert.fail("expected AbstractMethodError");
+        } catch (AbstractMethodError e) {
+            Assert.assertEquals("Invalid call to Copier.init(Run src, Run dst, FilePath, FilePath), passing an AbstractBuild " +
+                    "instance for the 'dst' arg when hudson.plugins.copyartifact.CopierTest$C3 does not implement the deprecated " +
+                    "version of 'init' that takes an AbstractBuild. Please supply a Run instance for the 'dst' arg.", e.getMessage());
```
jglick: This works, though it would probably suffice to

    assertTrue(e.getMessage(), e.getMessage().contains(C3.class.getName()));

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Rule
+    public JenkinsRule jenkinsRule = new JenkinsRule();
+
+    @Test
+    public void test_simpleUntriggeredCopy() throws Exception {
+        // create "project_1" with an archived artifact...
+        WorkflowJob project_1 = createWorkflow("project_1",
+                "sh 'echo hello > hello.txt'; " +
+                "step([$class: 'ArtifactArchiver', artifacts: 'hello.txt', fingerprint: true])");
```
jglick: BTW can also use the `archive` step, which is a little more concise, but otherwise the same.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        // Create "project_2", which triggers build of "project_1"
+        WorkflowJob project_2 = createWorkflow("project_2",
+                "build ('project_1'); " +
+                "step ([$class: 'CopyArtifact', projectName: 'project_1', filter: 'hello.txt']); " +
+                "step ([$class: 'ArtifactArchiver', artifacts: 'hello.txt', fingerprint: true]);");
+
+        // run project_2 and check that was able to copy from project_1
+        WorkflowRun b = jenkinsRule.assertBuildStatusSuccess(project_2.scheduleBuild2(0));
+        assertArtifactInArchive(b);
```
jglick: Is this actually testing anything distinct from `test_simpleUntriggeredCopy`? In either case, you are copying from the latest build of some other project. The fact that one triggered the other should not matter unless you are using some custom `Selector` which is picking an upstream build based on `UpstreamCause` or the like.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Test
+    public void test_downstreamCopyFromUpstream() throws Exception {
+        // create "project_1" with an archived artifact...
+        WorkflowJob project_1 = createWorkflow("project_1",
+                "sh 'echo hello > hello.txt'; " +
+                "step([$class: 'ArtifactArchiver', artifacts: 'hello.txt']); " +
+                "build ('project_2'); ");
```
jglick: Again the fact that `project_1` triggered `project_2` seems to be irrelevant for this test.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        // Create "project_2", which triggers build of "project_1"
+        // I'm sure there's a sweeter way of passing in the build number via a parameter
+        WorkflowJob project_2 = createWorkflow("project_2",
+                "step ([$class: 'CopyArtifact', projectName: 'project_1', filter: 'hello.txt', selector: [$class: 'SpecificBuildSelector', buildNumber: '#1']]); " +
```
jglick: This is going through the last fallback behavior in `SpecificBuildSelector.getBuild`, that checking `Run.getDisplayName`. Pass `'1'` not `'#1'`.

(`DescribableHelper` does _not_ perform the implicit coercion necessary to pass `1`, the `Integer`. Cf. `DescribableHelperTest.mismatchedTypes`.)

src/main/java/hudson/plugins/copyartifact/Copier.java
tfennelly: Was part of original code and I was avoiding making unrelated changes as requested. Will fix of course.

jglick: This is a new overload though, so the right time to fix the existing mistake.

tfennelly: Correction ... I introduced a new `init` and should have fixed it. My bad !

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
tfennelly: Right, probably not needed. Have removed and am running tests.

While `canCopyArtifact` can take `Job` for both of its args, I had in my head (wrongly I'm sure) that the original implementor only wanted to call that method for `AbstractProject` types (for some reason) and otherwise move on to the code after. In hindsight that doesn't seem to make sense.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
tfennelly: Removed.

src/test/java/hudson/plugins/copyartifact/CopierTest.java
tfennelly: If it's all the same, I'd prefer to leave as is.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
tfennelly: It's true that from a purely `CopyArtifact` perspective, the steps should be doing the same thing.
My thinking was that it verifies that `CopyArtifact` works in tandem with the `build` step (workflow `A` fires workflow `B`, waits for it to complete and then copies an artifact from it), which is not invalid and not a bad test imo.

That said, I'll remove if you prefer it not to be there.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
tfennelly: Testing that a workflow can successfully copy an artifact from a build that triggered it is not relevant?

tfennelly:
@jglick Jesse, let me know if you want me to remove those tests from `CopyArtifactWorkflowTest`.

src/main/java/hudson/plugins/copyartifact/Copier.java
tfennelly: @jglick interesting one here if I introduce the generic types on `Run src` ... it results in a compile time error re ambiguous method signatures. I need to read the Java language spec now lol

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
jglick: You have already tested that a workflow can copy an artifact from a build that did _not_ trigger it. The test makes no use of the triggering relationship. Meaning the test is harmless (other than time taken in `mvn test`), but redundant.

(This plugin _does_ have a selector specifically to pick the right upstream build. It cannot work for Workflows currently, because the whole apparatus of identifying upstream/downstream builds based on fingerprints and the dependency graph is so far restricted to `AbstractProject/Build`. If that limitation were removed in a subsequent version of core, then you could modify this plugin to let that selector deal with an arbitrary `Run`. To be meaningful, this test would then need to use that selector, _and_ run multiple upstream builds to verify that the artifact copied was the one from the upstream build triggering this downstream build, as opposed to simply picking the latest upstream build at the time the copy step happened to run.)

jglick:
:+1: from me, assuming the tests still pass of course. @ikedam?

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
tfennelly: I'll remove the test.

As an aside ... the `build` step takes named parameters, right? Would it not be possible to pass the build number as a parameter, therefore allowing the exact build to be copied from?

jglick:
Thus currently testing only with the default `StatusBuildSelector`, which is probably fine unless there is a reason to suspect problems with others. :+1:

ikedam:
Looks perfect!
I'm sorry not to have joined the review at all.

Though I planned to release this feature as a part of copyartifact-2.0 ([JENKINS-24892](https://issues.jenkins-ci.org/browse/JENKINS-24892])), I now plan to release this as copyartifact-1.34 as:
* I still don't find time to work for copyartifact-2.0.
* This keeps the backward compatibility.
* Many users should expect and wait for this feature.

Do you have other changes (pull requests) for the next release?

tfennelly:
Thanks @ikedam. When are you planning the 1.34 release for?

ikedam:
I'll make a new release tomorrow as there's no further change planned for 1.X.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -257,91 +320,102 @@ public boolean isFingerprintArtifacts() {
...
-            throws InterruptedException, IOException {
-        upgradeIfNecessary(build.getProject());
+    public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener) throws InterruptedException, IOException {
+        if (build instanceof AbstractBuild) {
+            upgradeIfNecessary(((AbstractBuild)build).getProject());
+        }
+
+        EnvVars env;
+        if (build instanceof AbstractBuild) {
+            env = build.getEnvironment(listener);
```
jglick: Wrong: [JENKINS-30357](https://issues.jenkins-ci.org/browse/JENKINS-30357) (should be lifted outside `if`-statement)

PR #57, opened 2015-02-06
[JENKINS-20546] Preserve symlinks when copying artifacts

jglick:
[JENKINS-20546](https://issues.jenkins-ci.org/browse/JENKINS-20546)

### DIFF
--- a/src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
+++ b/src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
@@ -85,7 +85,7 @@ import com.cloudbees.hudson.plugins.folder.Folder;
 import com.gargoylesoftware.htmlunit.FailingHttpStatusCodeException;
 import com.google.common.collect.Sets;
 
-import static org.junit.Assert.assertTrue;
+import org.jvnet.hudson.test.TestBuilder;
 
 /*
 // classes used only in testQueueItemAuthenticator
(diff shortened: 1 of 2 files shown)


jglick:
@reviewbybees

jglick:
Note: could be backported to 1.33.1-SNAPSHOT but the test fails in the setup because `FilePath.copyRecursiveTo` did not handle symlinks in 1.509 and throws an exception while archiving from the upstream project. Updating the parent to 1.532.3 avoids the exception, but the symlinks are skipped so the test still fails later because they are just absent from the downstream project. Updating the parent to 1.554.3 (or 1.554) lets the test pass.

recampbell:
:+1:

src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
```diff
@@ -1662,6 +1662,28 @@ public void testFilePermission() throws Exception {
             assertEquals(0755, w.child("artifactWithExecuteInSubdir.txt").mode() & 0777);
         }
     }
+
+    @Bug(20546)
```
tfennelly: Is `Bug` not deprecated?

tfennelly:
:+1:

src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
jglick: Yes but using it here makes backporting slightly simpler.

ikedam:
+1

I don't think we need to backport this to 1.33 as this isn't a critical issue and there are workarounds (like packing files with tar command).
Backporting is required only for security fixes.

ikedam:
Notes
* On Windows, tests need administrative privilege to pass. Windows requires administrative privilege to create symbolic links.
* The test requires Java >= 7. `FilePath#readLink` works only with Java >= 7.
    * I always use Java6 to make a release as I faced a problem that occurred only when compiled with Java6 ([JENKINS-21274](https://issues.jenkins-ci.org/browse/JENKINS-21274)). I used Java7 to release copyartifact 1.35.
    * (That should work even compiled with Java6 only if running with Java>=7)

PR #59, opened 2015-02-14
[JENKINS-26694] Make ParameterBuildFilter applicable to WorkflowJobs.

ikedam:
https://issues.jenkins-ci.org/browse/JENKINS-26694

`ParametersBuildFilter` doesn't work for `WorkflowJob`s as it extract parameters of upstream builds with `Run#getEnvironemnt(TaskListener)`.
You can extract build parameters from environments only in `AbstractBuild`s, and `WorkflowJob` is a subclass of `Run`, not `AbstractBuild`.

This is a workaround for that problem.

PR #61, opened 2015-03-03
[WiP] minor changes for Workflow support

ndeloof:
Let user rely on snippet generator to use copyartifact build step

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -512,7 +512,7 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
             else if (value.indexOf('$') >= 0)
                 result = FormValidation.warning(Messages.CopyArtifact_ParameterizedName());
             else {
-                AbstractProject nearest = AbstractProject.findNearest(value);
+                Job nearest = Items.findNearest(Job.class, value, parent);
                 String alternative = nearest != null ? nearest.getRelativeNameFrom(parent) : "?";
                 result = FormValidation.error(
                     hudson.tasks.Messages.BuildTrigger_NoSuchProject(
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -506,10 +506,14 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
                 }
             else if (value.indexOf('$') >= 0)
                 result = FormValidation.warning(Messages.CopyArtifact_ParameterizedName());
-            else
+            else {
+                AbstractProject nearest = AbstractProject.findNearest(value);
+                String alternative = nearest != null ? nearest.getRelativeNameFrom(parent) : "?";
                 result = FormValidation.error(
                     hudson.tasks.Messages.BuildTrigger_NoSuchProject(
-                        value, AbstractProject.findNearest(value).getName()));
+                            value, alternative));
+
+            }
             return result;
         }
 
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -132,7 +132,7 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
         // check the permissions only if we can
         StaplerRequest req = Stapler.getCurrentRequest();
         if (req!=null) {
-            AbstractProject<?,?> p = req.findAncestorObject(AbstractProject.class);
+            Job p = req.findAncestorObject(Job.class);
             if (p != null) {
                 ItemGroup<?> context = p.getParent();
 
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -493,11 +493,9 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
     public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
 
         public FormValidation doCheckProjectName(
-                @AncestorInPath AbstractProject anc, @QueryParameter String value) {
-            // Require CONFIGURE permission on this project
-            if (!anc.hasPermission(Item.CONFIGURE)) return FormValidation.ok();
+                @AncestorInPath ItemGroup parent, @QueryParameter String value) {
             FormValidation result;
-            Item item = Jenkins.getInstance().getItem(value, anc.getParent());
+            Item item = Jenkins.getInstance().getItem(value, parent);
             if (item != null)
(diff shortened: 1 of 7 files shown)


ndeloof:
@reviewbybees

ndeloof:
not sure why CI fails a test since https://jenkins.ci.cloudbees.com/job/plugins/job/copyartifact-plugin/151/ as previous build https://jenkins.ci.cloudbees.com/job/plugins/job/copyartifact-plugin/150/ was on same commit (maven release)

jglick:
I guess this is intended to follow up #55 and #59?

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -493,11 +498,9 @@ private boolean perform(Run src, Run<?,?> dst, String expandedFilter, String exp
     public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
 
         public FormValidation doCheckProjectName(
-                @AncestorInPath AbstractProject anc, @QueryParameter String value) {
-            // Require CONFIGURE permission on this project
-            if (!anc.hasPermission(Item.CONFIGURE)) return FormValidation.ok();
```
jglick: This would be a security regression. You need to retain `@AncestorInPath Job job` so you can still do a permission check on it.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -508,10 +511,14 @@ public FormValidation doCheckProjectName(
                 }
             else if (value.indexOf('$') >= 0)
                 result = FormValidation.warning(Messages.CopyArtifact_ParameterizedName());
-            else
+            else {
+                AbstractProject nearest = AbstractProject.findNearest(value);
+                String alternative = nearest != null ? nearest.getRelativeNameFrom(parent) : "?";
```
jglick: `Items.findNearest`

src/main/resources/hudson/plugins/copyartifact/CopyArtifact/config.jelly
```diff
@@ -26,7 +26,7 @@ THE SOFTWARE.
   <f:entry title="${%Project name}" field="projectName">
     <f:editableComboBox items="${app.topLevelItemNames}" clazz="setting-input"/>
   </f:entry>
-  <ca:selectorList currentSelector="${instance.buildSelector}"
+  <ca:selectorList currentSelector="${instance.selector}"
       name="selector" title="${%Which build}"/>
```
jglick: Would be nice to fix `selectorList.jelly` to take a `field`.

jglick:
Suggestion: if you add

```xml
      <dependency>
        <artifactId>workflow-step-api</artifactId>
        <version>${workflow.version}</version>
        <classifier>tests</classifier>
        <scope>test</scope>
      </dependency>
```

then it should be possible to use `StepConfigTester` on a `new CoreStep(new CopyArtifact(…))` to actually reproduce the `buildSelector` and `stableOnly` problems and verify the fix (as well as defending against some future regressions).

jglick:
Evaluate the [`CopyArtifactTest.testFieldValidation` failure](https://jenkins.ci.cloudbees.com/job/plugins/job/copyartifact-plugin/153/org.jenkins-ci.plugins$copyartifact/testReport/hudson.plugins.copyartifact/CopyArtifactTest/testFieldValidation/).

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
ndeloof: This permission check applies to current job configuration, not the path to job the method is going to validate. This doesn't make much sense, can be arbitrary job on jenkins. 
I've looked at git history to understand the intent but was imported with first commit.

ndeloof:
@reviewbybees ping

ikedam:
Let's make it clear what this request does.

You can test the behavior of the snippet generator in a following way:
* Create a new workflow project
* Select "Groovy CPS DSL" in Definition for Workflow
* Check "Snippet generator"
* Select "General Build Step" for "Sample Step"
* Select "Copy artifacts from another project" for "Build Step"
* Configure copyartifact, and click "Generate Groovy"

Result before this fix:
* The validation for "Project name" fails with NPE.
* The generated result is
```
step <object of type hudson.plugins.copyartifact.CopyArtifact>
```

Result after this fix:
```
step([$class: 'CopyArtifact', excludes: '', filter: '**/*', fingerprintArtifacts: true, flatten: false, optional: false, projectName: 'test', selector: [$class: 'StatusBuildSelector', stable: false], target: ''])
```

What this request does:
* Makes `CopyArtifact.DescriptorImpl#doCheckProjectName` work with `WorkflowJob` which is a sub class not of `AbstractProject` but of `Job`.
* Make getter names match the names of parameters in the constructor (or maybe setters with `DataBoundSetter`).

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
ikedam: What you want really do is to make `doCheckProjectName` applicable to `WorkflowJob`.
You can do that simply by replacing `AbstractProject` with `Job`.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
ikedam: +1 for `Items.findNearest`.

This change is supposed to avoid NPE when there are only `WorkflowJob`s and no `AbstractProject`s.
`Items.findNearest` enables us to display also a `WorkflowJob` as a candidate.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: > This permission check applies to current job configuration

You misunderstand completely. The intention is to avoid doing potentially sensitive computations unless the user is actually configuring _this_ (downstream) job. Otherwise someone can fake up a `…/checkProjectName` request in cURL and get some information they should not have access to based on job permissions.

As @ikedam says, just replace `AbstractProject` with `Job`.

jglick:
What is the status of this? It is not currently mergeable.

ikedam:
copyartifact-1.35.1 should work with workflow's snippet generator. (fixed with #62, #63).
It will be available in a day.
Please try that.

jglick:
The substitution of `AbstractProject` with `Job` in a couple of places is still valid, right?

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -132,7 +132,7 @@ public CopyArtifact(String projectName) {
         // check the permissions only if we can
         StaplerRequest req = Stapler.getCurrentRequest();
         if (req!=null) {
-            AbstractProject<?,?> p = req.findAncestorObject(AbstractProject.class);
+            Job p = req.findAncestorObject(Job.class);
```
ikedam: This (and only this) change isn't applied in 1.35.1 as @jglick points.
[JENKINS-28247](https://issues.jenkins-ci.org/browse/JENKINS-28247) may be caused for that.

PR #62, opened 2015-04-20
[JENKINS-28011] Using the standard f:dropdownDescriptorSelector

jglick:
[JENKINS-28011](https://issues.jenkins-ci.org/browse/JENKINS-28011)

Code simplification which offers an alternative fix to https://github.com/jenkinsci/jenkins/pull/1663.

I did not delete `selectorList.jelly` since it is still being used by `BuildSelectorParameter/config.jelly` and `index.jelly`, though that could surely be made to use the standard control as well.

Makes `CopyArtifactTest.testPermission` pass against 1.610. There are various other test failures running against this baseline, which I did not try to evaluate.

@reviewbybees


jglick:
Test failures because of the use of the nonstandard `selector` rather than the standard `instance` and databinding. Really the databinding in this plugin is a mess.

stephenc:
:+1:

ikedam:
Looks good.
As long as I tested, JENKINS-28011 also happens with `BuildSelectorParameter` (assumed "various other test failures running against 1.610").
We should remove `ca:selectorList` completely (even though JENKINS-28011 will be fixed in Jenkins-1.611).
I work for that in [JENKINS-28094](https://issues.jenkins-ci.org/browse/JENKINS-28094).

PR #63, opened 2015-04-25
Removed ca:selectorList

ikedam:
[JENKINS-28094](https://issues.jenkins-ci.org/browse/JENKINS-28094).
Please see also [JENKINS-28011](https://issues.jenkins-ci.org/browse/JENKINS-28011), #62.
This is an additional change for #62.

Replaces  `ca:selectorList` with `f:dropdownDescriptorSelector` provided in Jenkins core.
Now `ca:selectorList` is no longer used, and removed in this change.

This request also includes changes for [JENKINS-28096](https://issues.jenkins-ci.org/browse/JENKINS-28096), #61, as #62 outdated #61.


src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -497,7 +497,7 @@ private boolean perform(Run src, Run<?,?> dst, String expandedFilter, String exp
     public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
 
         public FormValidation doCheckProjectName(
-                @AncestorInPath AbstractProject anc, @QueryParameter String value) {
+                @AncestorInPath Job<?,?> anc, @QueryParameter String value) {
```
ikedam: A change for [JENKINS-28096](https://issues.jenkins-ci.org/browse/JENKINS-28096).
As workflow is a subclass of `Job`, validations for project names fails (cause an exception) when copyartifact used in workflow's snippet generator.

src/main/resources/hudson/plugins/copyartifact/BuildSelectorParameter/config.jelly
```diff
@@ -22,12 +22,12 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form" xmlns:ca="/hudson/plugins/copyartifact">
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form">
   <f:entry title="${%Name}" field="name" help="/help/parameter/name.html">
     <f:textbox/>
   </f:entry>
-  <ca:selectorList currentSelector="${instance.defaultSelector}"
-      name="defaultSelector" title="${%Default Selector}" omit="ParameterizedBuildSelector"/>
+  <f:dropdownDescriptorSelector field="defaultSelector"  help="/plugin/copyartifact/help-buildSelector.html"
+      title="${%Default Selector}" descriptors="${descriptor.availableBuildSelectorList}" />
```
ikedam: /plugin/copyartifact/help-buildSelector.html doesn't exist actually. strange.

ikedam:
@jglick @reviewbybees I want your reviews as this is an additional change for #62.

src/main/resources/hudson/plugins/copyartifact/BuildSelectorParameter/config.jelly
jglick: I may have moved it in #62 to be in the default location for `CopyArtifact/config.jelly`. Try just creating a `help-defaultSelector.html` in this directory, and deleting the `help` attribute.

src/main/resources/hudson/plugins/copyartifact/BuildSelectorParameter/index.jelly
```diff
@@ -22,10 +22,13 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
 -->
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:ca="/hudson/plugins/copyartifact">
-  <ca:selectorList currentSelector="${it.defaultSelector}" descriptor="${it.descriptor}"
-      name="parameter" title="${it.name}" description="${it.description}"
-      omit="ParameterizedBuildSelector">
-    <tr><td><input type="hidden" name="name" value="${it.name}"/></td></tr>
-  </ca:selectorList>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form">
+  <f:dropdownDescriptorSelector field="parameter" help="/plugin/copyartifact/help-buildSelector.html"
```
jglick: ditto

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: Right, this should help with the excepton.

The bogus object is something else, typically caused by some nonstandard databinding idiom, missing getter, etc.; but possibly a bug in Workflow in terms of handling unusual collection signatures or the like. You can add a `test`-scoped dependency on `workflow-api` and call `DescribableHelper.uninstantiate` on a configured `CopyArtifact` to see what went wrong. (Look at `DescribableHelperTest` for examples of the quasi-JSON expected results. There should be no Java objects other than primitive wrappers, `String`, `List`, and `Map`.)

jglick:
:+1: other than the `help` oddness.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
ikedam: I think you're talking about the latter issue described in [JENKINS-28096](https://issues.jenkins-ci.org/browse/JENKINS-28096).
That's already fixed with your change #62 . Thanks.
I updated [JENKINS-28096](https://issues.jenkins-ci.org/browse/JENKINS-28096).

src/main/resources/hudson/plugins/copyartifact/BuildSelectorParameter/config.jelly
ikedam: Oh, it was not only moved, but renamed. That's why my `find` command failed to find it.
I'll add help-defaultSelector.html. Thanks.

ikedam: I considered it again, and found that it would result creating three copies of help-selector.html.
I'll add a commit to revert the move of help-defaultSelector.html in #62.

ikedam: Unfortunately, `dropdownDescriptorSelector` doesn't support `help` attribute... It always decide help files to use by passed descriptors.
I'll create three copies of help-selector.html.

src/main/java/hudson/plugins/copyartifact/BuildSelectorParameter.java
```diff
@@ -90,6 +97,35 @@ public String getDisplayName() {
...
+                // for `defaultSelector` ("Default Selector" field) in project configuration pages
+                // and the value of build parameter ("Build selector for Copy Artifact" field)
+                // in "This build requires parameters" pages.
+                Descriptor<?> d = Jenkins.getInstance().getDescriptor(CopyArtifact.class);
+                if (d != null) {
+                    return d.getHelpFile("selector");
+                }
+            }
+            return super.getHelpFile(fieldName);
+        }
```
ikedam: This displays the same help for `Copyartifact#getSelector` and `BuildSelectorParameter#getDefaultSelector` and the value in "The build requires parameters" page.

PR #65, opened 2015-05-10
fixes JENKINS-18938 when setting COPYARTIFACT_BUILD_NUMBER_<job> don'…

GeekOffTheStreet:
…t replace numbers in job name

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -604,7 +604,7 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
                 return;
             }
             data.put("COPYARTIFACT_BUILD_NUMBER_"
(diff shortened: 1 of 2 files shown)


ikedam:
I agree the current behavior is bad.
But I don't think to merge this request as:
* This change might break existing setups.
* The mapping from project names to variable names are complicated and it's not easy for users to know the variable name they need, even with this change.

How about adding a new field to specify the variable name to hold the build number?

jdmears2:
Without a doubt, the change is not backwards compatible for anyone currently utilizing the variable for a job with a number in it.

With respect to your second bullet, yes the renaming is confusing.  However, this change would put the behavior of this plugin in line with the parameterized build plugin: https://wiki.jenkins-ci.org/display/JENKINS/Parameterized+Trigger+Plugin

I'm not sure I understand your last sentence.  Are you suggesting a completely new environment variable for each job?

ikedam:
Like this.
copyartifact-buildnumbervariable

ikedam:
merged #69 instead.

PR #66, opened 2015-06-28
[JENKINS-28972] Make SimpleBuildSelectorDescriptor usable from other plugins

ikedam:
[JENKINS-28972](https://issues.jenkins-ci.org/browse/JENKINS-28972).

Copyartifact-1.35.1 gets to use {{f:dropdownDescriptorSelector}}.
It makes {{SimpleBuildSelectorDescriptor}} unusable from other plugins for the Jenkins core issue [JENKINS-29048](https://issues.jenkins-ci.org/browse/JENKINS-29048).

This should finally be fixed in Jenkins core, but it doesn't make copyartifact work with current Jenkins cores.
This is a workaround for that.


ikedam:
This problem should cause 500 errors, but the test causes JavaScript errors.
I'm not sure why, but anyway my change fixed that JavaScript errors and the test passes.

PR #67, opened 2015-07-08
Configured FindBugs xml output

rsandell:
So Jenkins can pick them up

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,6 +95,16 @@
                 <artifactId>maven-hpi-plugin</artifactId>
                 <version>1.95</version>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>findbugs-maven-plugin</artifactId>
+                <version>3.0.1</version>
+                <configuration>
+                    <xmlOutput>true</xmlOutput>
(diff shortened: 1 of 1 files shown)


rsandell:
25 issues found in my initial sweep, too many to fix in my quick drive-by.

oleg-nenashev:
:bee: MOAR findbugs. Was the code clear?

ikedam:
Why didn't Jenkins pick up these 25 issues?

rsandell:
@oleg-nenashev in some places it was clear. But since there was so many I didn't look at them all.

rsandell:
@ikedam I haven't configured the job to run findbugs yet. I wanted to get this in first.

stephenc:
:bee:

rsandell:
@reviewbybees done

rsandell:
@ikedam Are you maintaining this plugin as well? Would it be ok to merge this in so we can start track the findbugs trends, and then when we are down do 0 we could turn on the failing of new warnings.

ikedam:
OK, builds won't fail for existing findbugs warnings, will they?

ikedam:
@rsandell 
Please turn on findbugs feature on Jenkins, and trigger builds twice (to see existing find bugs warnings doesn't cause build failures).

PR #69, opened 2015-08-10
[JENKINS-18938] Added a field to specify the suffix for the variable to store the build number

ikedam:
[JENKINS-18938](https://issues.jenkins-ci.org/browse/JENKINS-18938)
Copyartifact stores the number of the build where copied artifacts to a variable "COPYARTIFACT_BUILD_NUMBER_(PROJECTNAME)".
Not all project names are usable for variable names and copyartifact substitutes invalid letters in project names.
That contains following problems;
* It's difficult for users to know the actual variable name where the build number is stored.
    * Especially when the source project is in different folder.
* "Invalid letters" contain numeric letters even though they are actually usable for variable names.

This change introduces a new field "Result variable suffix" which specified the suffix for "COPYARTIFACT_BUILD_NUMBER_(SUFFIX)".

* It is located in "Advanced" section.
    resultvariablesuffix-1
* You can specify the suffix for the variable name. 
resultvariablesuffix-2

If nothing is specified or invalid one is specified, the project name is used as the suffix just like the existing behavior.
This keeps the backward compatibility.

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -48,6 +48,7 @@ import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.Builder;
 import hudson.util.DescribableList;
 import hudson.util.FormValidation;
+import hudson.util.VariableResolver;
 import hudson.util.XStream2;
 
 import java.io.IOException;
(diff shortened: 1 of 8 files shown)


daniel-beck:
Why not specify the full variable name?

ikedam:
Because other variables like `COPYARTIFACT_PROJECTNAME_(SUFFIX)` may be supported in the future.

PR #72, opened 2015-10-12
Add missing credentials dependency

kwhetstone:
The tests for any class using FolderCredentials was failing with a
ClassDefNotFound.  This happens because the credentials plugin is
"optional" for the cloudbees-folder plugin (which is used by this class
for testing only).  The credentials plugin also has to be included in
the pom for test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,12 @@
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>credentials</artifactId>
+        <version>1.8.3</version>
+        <scope>test</scope>
+      </dependency>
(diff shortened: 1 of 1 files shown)


ikedam:
Unnecessary change as tests in copyartifact-plugin doesn't use `FolderCredentialsProvider`.
All tests works fine even without credentials-plugin as far as I know.
Would you let me know why you need this change?

kwhetstone:
Hey, so I'm running plugin-compat-tester with errors, and this error was being thrown all over the place.  It turns out there is a different problem in the tests with this plugin; running into a nullpointer exception with one of the tests.  I'll close this pull request.

PR #73, opened 2015-10-15
Add option to 'Upstream build that triggered this job' to allow upstream dependencies

gvasko:
The CopyArtifact plugin with 'Upstream build that triggered this job' works fine for us for a chain of jobs when the build started with the first job that triggers the next one, which triggers the next, and so on. The advantage of the Upstream-build strategy is that it works even with overlapping chains of builds. But there is a problem when we have overlapping chains and the build is started with an intermediate job.

Let's consider a simple scenario: each job triggers the next one
ModuleA --> ModuleB --> Publisher
where 
ModuleB depends on artifacts of ModuleA (using the Upstream-build strategy),
Publisher copies artifacts from ModuleA and ModuleB (using the Upstream-build strategy),
ModuleB needs significantly more time to build than ModuleA, so ModuleA can build while a ModuleB build is running.

Example execution:
ModuleA#1 is triggered by SCM, builds successfully.
ModuleB#1 is triggered by ModuleA#1, copies its artifacts and builds successfully.
Publisher#1 is triggered by ModuleB#1, copies artifacts from ModuleA#1 and ModuleB#1.
ModuleB#2 is triggered by SCM, copies artifacts from ModuleA#1 (fallback to last successful), build started...
ModuleA#2 is triggered by SCM, builds successfully.
... ModuleB#2 finished.
Publisher#2 is triggered by ModuleB#2, copies artifacts from ModuleB#2 and ModuleA#2 (fallback) instead of ModuleA#1, which ModuleB#2 depends on.
ModuleB#3 and Publisher#3 works as expected.
    
This scenario happens more easily if the chain is longer and contains more than 3 jobs.
The problem here is the inconsistent builds (here the Publisher), or the violated isolation of build chains.

I hope what I've written so far is a reasonable issue.

What I found to this is that, the Upstream-build strategy in CopyArtifact plugin could include the upstream builds of the target build, not just the builds that triggered it. I have implemented this change introducing an additional checkbox for this, tried on a local Jenkins installation and seems working.

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/TriggeredBuildSelector.java
+++ b/src/main/java/hudson/plugins/copyartifact/TriggeredBuildSelector.java
@@ -95,13 +95,13 @@ public class TriggeredBuildSelector extends BuildSelector {
     };
     private Boolean fallbackToLastSuccessful;
     private final UpstreamFilterStrategy upstreamFilterStrategy;
-    private Boolean allowUpstreamDependencies;
+    private boolean allowUpstreamDependencies;
 
     @DataBoundConstructor
     public TriggeredBuildSelector(boolean fallbackToLastSuccessful, UpstreamFilterStrategy upstreamFilterStrategy, boolean allowUpstreamDependencies) {
         this.fallbackToLastSuccessful = fallbackToLastSuccessful ? Boolean.TRUE : null;
         this.upstreamFilterStrategy = upstreamFilterStrategy;
-        this.allowUpstreamDependencies = allowUpstreamDependencies ? Boolean.TRUE : null;
+        this.allowUpstreamDependencies = allowUpstreamDependencies;
     }
 
     @Deprecated
@@ -143,7 +143,7 @@ public class TriggeredBuildSelector extends BuildSelector {
     }
     
     public boolean isAllowUpstreamDependencies() {
-        return allowUpstreamDependencies != null && allowUpstreamDependencies.booleanValue();
+        return allowUpstreamDependencies;
     }
     
     @Override
--- a/src/main/java/hudson/plugins/copyartifact/TriggeredBuildSelector.java
+++ b/src/main/java/hudson/plugins/copyartifact/TriggeredBuildSelector.java
@@ -27,12 +27,14 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 import jenkins.model.Jenkins;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.model.Result;
 import hudson.model.AbstractProject;
+import hudson.model.AbstractBuild;
 import hudson.model.Cause;
 import hudson.model.Cause.UpstreamCause;
 import hudson.model.Job;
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/plugins/copyartifact/TriggeredBuildSelector.java
```diff
@@ -93,16 +95,18 @@ public boolean isForGlobalSetting() {
     };
     private Boolean fallbackToLastSuccessful;
     private final UpstreamFilterStrategy upstreamFilterStrategy;
+    private Boolean allowUpstreamDependencies;
```
ikedam: You can use `boolean` here as it's set to `false` if not configured.
You have to use `Boolean` if:
* You want to know it's not configured and want to run an upgrading process.
* You want to set it to `true` if not configured.

src/main/java/hudson/plugins/copyartifact/TriggeredBuildSelector.java
```diff
@@ -93,16 +95,18 @@ public boolean isForGlobalSetting() {
...
 
     @DataBoundConstructor
-    public TriggeredBuildSelector(boolean fallbackToLastSuccessful, UpstreamFilterStrategy upstreamFilterStrategy) {
+    public TriggeredBuildSelector(boolean fallbackToLastSuccessful, UpstreamFilterStrategy upstreamFilterStrategy, boolean allowUpstreamDependencies) {
```
ikedam: Please add an overloaded constructor to keep comparibility (like `TriggerredBuildSelector(boolean)`).

src/test/java/hudson/plugins/copyartifact/DownstreamBuildSelectorTest.java
```diff
@@ -31,7 +31,7 @@
 import java.util.Map;
 import java.util.Set;
 
-import hudson.plugins.copyartifact.testutils.CopyArtifactUtil;
+import hudson.plugins.copyartifact.testutils.CopyArtifactUtil;
```
ikedam: Note: This changes the line break from LF to CR+LF.
The root cause is inconsistent line breaks introduced in #55.
(I couldn't find them in the review with github web interfaces...)

I'll accept this change even though pull requests would better not to contain unrelated changes, as:
* Not every developers use editors that can handle files with inconsistent line breaks.
* This should be fixed anyway.

src/main/java/hudson/plugins/copyartifact/TriggeredBuildSelector.java
```diff
@@ -138,69 +142,61 @@ public boolean isUseNewest() {
...
+                    upstreamBuilds.add(upstreamRun);
+                }
+            }
+        }
+
+        if (isAllowUpstreamDependencies() && (parent instanceof AbstractBuild)) {
+            AbstractBuild<?, ?> parentBuild = (AbstractBuild<?,?>)parent;
+            
+            Map<AbstractProject, Integer> parentUpstreamBuilds = parentBuild.getUpstreamBuilds();
+            for (AbstractProject project : parentUpstreamBuilds.keySet()) {
```
ikedam: Please use `Map#enteySet`.

You can use findbugs to see those warnings.
Run `mvn findbugs:findbugs` after compiling your codes.

gvasko:
Thank you, I will fix them soon.

src/test/java/hudson/plugins/copyartifact/DownstreamBuildSelectorTest.java
gvasko: As far as I can see, this was the only line in this file with LF line ending. I've found 26 files altogether that dos2unix would change.

ikedam: You can see changed lines in "Files changed" tabs and will find lines marked changed even they looks unchanged. Those lines were ended with LF.
Anyway, don't mind those changes as I described.

ikedam:
Thanks for the update!
It's time to merge.

gvasko:
Thanks for your support.

ikedam:
Released as copyartifact-1.37.
It will be available in the update center in a day.

gvasko:
Thanks, that sounds fine.

PR #75, opened 2015-11-16
Use run parameter

anentropic:
This is an attempt to revive and fix PR #17

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/ParameterizedBuildSelector.java
+++ b/src/main/java/hudson/plugins/copyartifact/ParameterizedBuildSelector.java
@@ -28,9 +28,12 @@ import java.util.logging.Logger;
 
 import hudson.EnvVars;
 import hudson.Extension;
+import hudson.model.ParameterValue;
 import hudson.model.Descriptor;
 import hudson.model.Job;
+import hudson.model.ParametersAction;
 import hudson.model.Run;
+import hudson.model.RunParameterValue;
 
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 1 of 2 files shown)


ikedam:
Link to JIRA ticket: https://issues.jenkins-ci.org/browse/JENKINS-21519

ikedam:
I think you can use `RunParameter` in this way:
* project: `${SELECTOR_JOBNAME}`
* selector: `SpecificBuildSelector`
* build: `${SELECTOR_NUMBER}`

Doesn't that work? (I haven't test it yet)

ikedam:
-1 for this change:
* `RunParameterValue` includes project names and it cause the project field of copyartifact ignores. It may allow bypassing permission checks.
* There's no good reason to extend `ParameterizedBuildSelector` instead introducing a new selector.

anentropic:
It does work with a 'String parameter' and giving the env variable to `SpecificBuildSelector`... however you lose the nice UI for picking a build number.

I don't understand what is the purpose of 'Run parameter' if you can't use it for something like this?

You'll notice in the original ticket https://issues.jenkins-ci.org/browse/JENKINS-12918  that there are several people who, like me, naturally assumed you could use a 'Run parameter' to select the build run, but instead receive a very obscure error message.

ikedam:
I don't think you know run parameter.
Have a look on the help: https://github.com/jenkinsci/jenkins/blob/master/war/src/main/webapp/help/parameter/run-project.html

anentropic:
I thought it was this one: https://github.com/jenkinsci/jenkins/blob/master/war/src/main/webapp/help/parameter/run.html

ikedam:
I tested and saw that my proposed way works as expected.
You can already use run parameters with specific build selectors and ${PARAMETER_NAME_NUMBER}, and no need to extend copyartifact.

anentropic:
Run parameter + specific build selector does not work:

`ERROR: Unable to find a build for artifact copy from: Unit-Tests`

The artifact exists. It works if I use string parameter, but I wanted to get the better UI selector.

PR #76, opened 2015-12-30
Do not allow httpclient/httpcore from maven-plugin to override jenkins-test-harness deps

jglick:
When run against 1.625.2, `DownstreamBuildSelectorTest.testConfiguration` fails with a JS NPE inside `confirm.js`

```javascript
if (btn.parentNode.parentNode.classList.contains("advanced-button")) {
```

Seems to be a regression from https://github.com/jenkinsci/jenkins/pull/1854 but only when using the old forked HtmlUnit.

This patch just ensures that we pick up some test-scope classes from the test harness, not the Maven plugin, so that we *can* run against a newer test harness (which then passes).

@reviewbybees esp. @kwhetstone

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,14 @@
             <groupId>org.jenkins-ci</groupId>
             <artifactId>SECURITY-144-compat</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
+          </exclusion>
         </exclusions>
       </dependency>
       <!-- Used for UI test -->


ikedam:
What do you mean with "run a test against 1.625.2"?
(I'm not sure whether this change requires a new release).

When I changed the version in pom.xml to 1.625.2 and running `mvn test -Dtest=DownstreamBuildSelectorTest#testConfiguration`:

* Switching the core to 1.625.2 also requires changes for `BuildSelectorParameterTest` (they are caused for jenkins-test-harness changed the version of htmlunit):
    * `WebRequestSettings` is renamed to `WebRequest`.
    * `WebClient#setThrowExceptionOnFailingStatusCode` is moved to `WebClientOptions`.
    * htmlunit uses `com.gargoylesoftware.htmlunit.util.NameValuePair` instead of `org.apache.commons.httpclient.NameValuePair`.
* The test failed like this in my environment before merging this change (anyway, this change fixed the problem):
```
hudson.plugins.copyartifact.DownstreamBuildSelectorTest
testConfiguration(hudson.plugins.copyartifact.DownstreamBuildSelectorTest)  Time elapsed: 33.453 sec  <<< ERROR!
java.lang.NoClassDefFoundError: org/apache/http/cookie/CookieSpecProvider
	at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
	at com.gargoylesoftware.htmlunit.WebClient.createWebConnection(WebClient.java:1956)
	at com.gargoylesoftware.htmlunit.WebClient.init(WebClient.java:236)
	at com.gargoylesoftware.htmlunit.WebClient.<init>(WebClient.java:211)
	at org.jvnet.hudson.test.JenkinsRule$WebClient.<init>(JenkinsRule.java:1871)
	at org.jvnet.hudson.test.JenkinsRule.createWebClient(JenkinsRule.java:1856)
	at hudson.plugins.copyartifact.DownstreamBuildSelectorTest.testConfiguration(DownstreamBuildSelectorTest.java:100)
```

pom.xml
```diff
@@ -37,6 +37,14 @@
...
             <artifactId>SECURITY-144-compat</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
+          </exclusion>
```
ikedam: Would you add a short comment for those exclusions in pom.xml?

kwhetstone:
1.625.2+ changed to the new htmlunit classes which fixed some issues, but introduced some new ones.  The best way to run copyartifact tests now (against the current core) is to compile it against it's parent version and use that version of the jenkins-test-harness.  If going the route of releasing a new version, the other problems you found should be addressed; this cropped up in a few different tests, but luckily was all the same root cause.
:bee:

jglick:
Should have mentioned that the problem arises when using `plugin-compat-tester`, which updates the baseline in a working copy; this patch is to correct the `NoClassDefFoundError` you saw. It simply ensures that the versions of two libraries are picked up from the correct transitive dependency, rather than some old versions bundled in `maven-plugin`.

`BuildSelectorParameterTest` is another issue, corrected if you use the _older_ test harness (but then `DownstreamBuildSelectorTest` fails with the JS issue, apparently a bug in our HtmlUnit fork). The ultimate solution will need to be a standalone test harness.

tfennelly:
:bee:

jglick:
@reviewbybees done

ikedam:
Steps to reproduce the problem:
* Change the target Jenkins version to 1.625.2.
* Add following dependency
```
      <dependency>
        <groupId>org.jenkins-ci.main</groupId>
        <artifactId>jenkins-test-harness</artifactId>
        <version>1.580.1</version>
      </dependency>
```
* Run `mvn test -Dtest=DownstreamBuildSelectorTest#testConfiguration`


I couldn't reproduce the problem with Firefox accessing the Jenkins instance launched by the test.
And I concludes that this is a client issue, the problem doesn't occur in user environments, and this doesn't require a new release.

ikedam:
That problem doesn't look fixed even with this change.
I think this change can be merged as a preparation to upgrade the target Jenkins version in future, but I don't think I have got the actual purpose of this change.

`plugin-compat-tester` should fail compiling `BuildSelectorParameterTest`, shouldn't it?
Is there a running instance of `plugin-compat-tester`?

jglick:
> I concludes that this is a client issue, the problem doesn't occur in user environments

Exactly, it looks to be a bug in the JS interpreter used by our old HtmlUnit fork.

> `plugin-compat-tester` should fail compiling `BuildSelectorParameterTest`, shouldn't it?

If you use a new test harness, yes. I plan to propose that the test harness be independently versioned, so that plugins like this can use a new harness (adjusting tests to match) but an old core, without needing the somewhat ugly trick of overriding the harness dependency. So this is a preparatory patch, but I think it makes sense in isolation.

> Is there a running instance of `plugin-compat-tester`?

I think the job @kwhetstone was running is not public currently, as it is testing a mixture of OSS and proprietary plugins. If you are interested, I can supply CLI arguments to run just the tests of this plugin. But it is not terribly interesting—you can as easily tweak `pom.xml` and `mvn test`.

ikedam:
I got it, thanks.

ikedam:
Merged in #77.

PR #77, opened 2016-01-06
Do not allow httpclient/httpcore from maven-plugin to override jenkins-test-harness deps

ikedam:
Added a comment for #76 ,dependency exclusions in pom.xml.


jglick:
:+1:

PR #78, opened 2016-01-14
[JENKINS-33389] Adapt to new parent POM and split JTH

andresrc:
[JENKINS-33389](https://issues.jenkins-ci.org/browse/JENKINS-33389)

* [x] Adapt to new parent POM
* [x] Adapt to split JTH.
* [x] Fix javadocs for Java 8
* [x] Correct line breaks in modified files
* [x] Migrate tests from `HudsonTestCase` to `JenkinsRule`.
* [x] Check PCT scenario against 1.642.3

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.3</version>
+        <version>2.5</version>
     </parent>
 
     <artifactId>copyartifact</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.580.1</version>
+        <version>2.3</version>
     </parent>
 
     <artifactId>copyartifact</artifactId>
@@ -22,7 +22,9 @@
     </licenses>
 
     <properties>
-        <workflow.version>1.1</workflow.version>
+      <jenkins.version>1.580.1</jenkins.version>
+      <java.level>6</java.level>
+      <workflow.version>1.1</workflow.version>
     </properties>
 
     <dependencies>
@@ -53,6 +55,12 @@
       </dependency>
       <!-- Used for UI test -->
       <dependency>
+        <groupId>org.jenkins-ci.main</groupId>
+        <artifactId>jenkins-test-harness-tools</artifactId>
+        <version>2.0</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>cloudbees-folder</artifactId>
         <version>4.0</version>
@@ -62,6 +70,24 @@
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>matrix-project</artifactId>
         <version>1.3</version>
+      </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>junit</artifactId>
+        <version>1.10</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>credentials</artifactId>
+        <version>1.24</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>matrix-auth</artifactId>
+        <version>1.2</version>
+        <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -99,31 +125,7 @@
             </exclusion>
         </exclusions>
       </dependency>
-      <!-- to suppress known (and harmless) findbugs issues -->
-      <dependency>
-          <groupId>com.google.code.findbugs</groupId>
-          <artifactId>annotations</artifactId>
-          <version>3.0.0</version>
-      </dependency>
     </dependencies>
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.jenkins-ci.tools</groupId>
-                <artifactId>maven-hpi-plugin</artifactId>
-                <version>1.95</version>
-            </plugin>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
-                <artifactId>findbugs-maven-plugin</artifactId>
-                <version>3.0.1</version>
-                <configuration>
-                    <xmlOutput>true</xmlOutput>
-                    <failOnError>false</failOnError>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
     <reporting>
       <plugins>
         <plugin>
@@ -147,19 +149,20 @@
       <tag>HEAD</tag>
   </scm>
 
-    <repositories>
-        <repository>
-            <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
-        </repository>
-    </repositories>
+  <repositories>
+    <repository>
+      <id>repo.jenkins-ci.org</id>
+      <url>http://repo.jenkins-ci.org/public/</url>
+    </repository>
+  </repositories>
 
-    <pluginRepositories>
-        <pluginRepository>
-            <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
-        </pluginRepository>
-    </pluginRepositories>
-</project>  
+  <pluginRepositories>
+    <pluginRepository>
+      <id>repo.jenkins-ci.org</id>
+      <url>http://repo.jenkins-ci.org/public/</url>
+    </pluginRepository>
+  </pluginRepositories>
+
+</project>
   
 
--- a/src/main/java/hudson/plugins/copyartifact/BuildSelectorParameter.java
+++ b/src/main/java/hudson/plugins/copyartifact/BuildSelectorParameter.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.logging.Logger;
 
+import com.thoughtworks.xstream.XStreamException;
 import jenkins.model.Jenkins;
 import hudson.DescriptorExtensionList;
 import hudson.Extension;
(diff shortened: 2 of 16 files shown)


pom.xml
```diff
@@ -22,7 +22,11 @@
     </licenses>
 
     <properties>
-        <workflow.version>1.1</workflow.version>
+      <jenkins.version>1.580.1</jenkins.version>
+      <!-- TODO check if we'd better add the PI. -->
+      <jelly.requirePI>false</jelly.requirePI>
```
jglick: Yes you should.

andresrc:
CI builds crashing (one was OOM, the other did not capture forked VM error). Tested locally.

ikedam:
JTH = Jenkins Test Harness

See also:
* [PROPOSAL] New Parent POM for Jenkins Plugins
    * https://groups.google.com/forum/#!msg/jenkinsci-dev/3_IMnXmtuDw/TEeowLK-BgAJ
    * https://github.com/jenkinsci/plugin-pom
* [PROPOSAL] Split of `jenkins-test-harness` out of `jenkinsci/jenkins`
    * https://groups.google.com/forum/#!msg/jenkinsci-dev/zviuh6z5_KM/OUKrvgvGAwAJ
    * https://github.com/jenkinsci/jenkins-test-harness
    * https://github.com/jenkinsci/jenkins-test-harness-tools

src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
```diff
@@ -73,14 +73,7 @@
...
 import org.junit.Test;
-import org.jvnet.hudson.test.Bug;
-import org.jvnet.hudson.test.ExtractResourceSCM;
-import org.jvnet.hudson.test.HudsonTestCase;
-import org.jvnet.hudson.test.Issue;
-import org.jvnet.hudson.test.CaptureEnvironmentBuilder;
-import org.jvnet.hudson.test.FailureBuilder;
-import org.jvnet.hudson.test.MockFolder;
-import org.jvnet.hudson.test.UnstableBuilder;
+import org.jvnet.hudson.test.*;
```
ikedam: I prefer specific imports.

ikedam:
Please turn off autoformatter of your IDE and don't make unrelated whitespace changes.
(even though there're inconsistent formattings or useless whitespaces)

ikedam:
Hmm... whitespace changes in `CopyartifactPermissionProperty.java` looks caused for existing inconsistent line breaks (LFs and CRLFs).
Those changes are acceptable as all editors cannot handle mixed line break characters in a file.

src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
andresrc: So do I, IDE misconfiguration in this project, will fix.

andresrc:
Yes, IDEA proposed to unify line breaks. I did not include it so as not to include unrelated changes. If it's acceptable I will include the changes.

andresrc:
Updated by force-pushing as it was easier to fix the line-breaks in the affected files.

pom.xml
```diff
@@ -64,6 +72,24 @@
         <version>1.3</version>
       </dependency>
       <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>junit</artifactId>
+        <version>1.10</version>
+        <scope>test</scope>
+      </dependency>
```
ikedam: (Note for me)
matrix-project depends on junit-plugin (via Jenkins core).
Jenkins core 1.580.1 bundles junit-1.0.

pom.xml
```diff
@@ -64,6 +72,24 @@
...
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>credentials</artifactId>
+        <version>1.24</version>
+        <scope>test</scope>
+      </dependency>
```
ikedam: (Note for me)
credentials-plugin is not necessary to have tests succeeds.
Though, it is useful to suppress error messages from folder-plugin.

pom.xml
```diff
@@ -64,6 +72,24 @@
...
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>matrix-auth</artifactId>
+        <version>1.2</version>
+        <scope>test</scope>
+      </dependency>
```
ikedam: (Note for me)
Some test codes depends on matrix-auth to configure authorizations.

ikedam:
`mvn test` also fails in my environment
* `CopyArtifactTest` fails with `OutOfMemory`
    * I had to add a following code to pom.xml to run tests.

        ```
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <configuration>
            <argLine>-Xmx1024m</argLine>
        </configuration>
      </plugin>
    </plugins>
  </build>
    ```
    * Jenkins-test-hanrness might have a memory leaking issue.
    * We might have to have `CopyArtifactTest` migrate to `JenkinsRule` (junit4)
* `PluginAutomaticTestBuilder$OtherTests#testPluginActive` failed with "Plugin copyartifact failed to start"
    * I have to see more details about that.

jglick:
I switched the build to run on Java 8, let us see if that helps.

pom.xml
jglick: You can use `MockAuthorizationStrategy` if you pick up a sufficiently new `jenkins-test-harness`.

src/main/java/hudson/plugins/copyartifact/BuildSelectorParameter.java
```diff
@@ -84,7 +85,10 @@ private StringParameterValue toStringValue(BuildSelector selector) {
 
     /**
      * Convert xml fragment into a BuildSelector object.
-     * @throws XStreamException or ClassCastException if input is invalid
+     * @param xml XML fragment to parse.
+     * @return the BuildSelector represented by the input XML.
+     * @throws XStreamException if the object cannot be deserialized
+     * @throws ClassCastException if input is invalid
```
jglick: You know you need not correct every Javadoc _warning_, right? Just _errors_.

src/test/java/hudson/plugins/copyartifact/CopyArtifactPermissionPropertyTest.java
```diff
@@ -269,4 +270,15 @@ public void testDescriptorDoAutoCompleteProjectNames() throws Exception {
...
     }
+
+    /**
+     * Creates an empty Matrix project with the provided name.
+     *
+     * @param name Project name.
+     * @return an empty Matrix project with the provided name.
+     */
+    private MatrixProject createMatrixProject(String name) throws IOException {
+        return j.jenkins.createProject(MatrixProject.class, name);
```
jglick: Clearer just to inline IMO.

src/test/java/hudson/plugins/copyartifact/TriggeredBuildSelectorTest.java
```diff
@@ -926,4 +924,16 @@ public void testMavenModule() throws Exception {
...
         );
     }
+
+    /**
+     * Creates an empty Maven project with an unique name.
+     *
+     * @return an empty Maven project with an unique name.
+     */
+    private MavenModuleSet createMavenProject() throws IOException {
+        MavenModuleSet mavenModuleSet = j.jenkins.createProject(MavenModuleSet.class, "test"+j.jenkins.getItems().size());
```
jglick: The autogeneration of project names is pretty stupid when every test case gets its own fresh Jenkins instance anyway. Simpler and clearer to just specify a name.

src/test/java/hudson/plugins/copyartifact/TriggeredBuildSelectorTest.java
```diff
@@ -926,4 +924,16 @@ public void testMavenModule() throws Exception {
...
+        mavenModuleSet.setRunHeadless(true);
```
jglick: Silly to be repeating this incantation in every test using `maven-plugin`. Better to just fix that plugin to set `runHeadless` to true by default when `Main.isUnitTest()`.

jglick: (or always)

jglick:
:bee:

ikedam:
Sorry for leaving this request for so long time... I can't have time to work for this (e.g. migrating CopyArtifactTest to junit4) yet.

@jglick 
Would you tell me why migrating to Java8 fixes `OutOfMomoryError`? (if you know)

ikedam:
Thanks for migrating tests to `JenkinsRule`.
Unfortunately, `OutOfMemoryError` still occurs when tested with Java7.
But I finally found it an existing potential issue (that is, this change doesn't cause that problem):

* Every time a test case run, the usage of PermGen increases.
    * I suppose PermGen is consumed when a plugin is loaded and a new class loader is created.
    * This occurs even with jenkins-test-harness provided by Jenkins 1.580.1.
* The initial PermGen usage of jenkins-test-harness >= 2.0 is larger than that of jenkins-test-harness < 2.0.
* That results `OutOfMemoryError` with jenkins-test-harness >= 2.0. Though, I believe I can trigger the same issue even with jenkins-test-harness < 2.0 just by adding more test cases.
    * `OutOfMemoryError` occurs only when running `mvn test`. `mvn test -Dtest=CopyartifactTest` doesn't cause `OutOfMemoryError` as "enough" test cases run only with `CopyArtifactTest`.
* `OutOfMemoryError` doesn't occur with Java8, as Java8 no longer uses PermGen.

I conclude that this change is good and ready to merge.
I plan to merge this change after other pull requests (#79, #81) merged as that may cause conflicts and requires manual merging.

src/test/java/hudson/plugins/copyartifact/TriggeredBuildSelectorTest.java
```diff
@@ -46,11 +47,8 @@
 
 import org.junit.Rule;
 import org.junit.Test;
-import org.jvnet.hudson.test.Bug;
-import org.jvnet.hudson.test.ExtractResourceSCM;
-import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.*;
```
ikedam: You'd better to configure your IDE not to use wildcard imports. It's often considered a bad style.

andresrc:
@ikedam Please notify when https://github.com/jenkinsci/copyartifact-plugin/pull/79 and https://github.com/jenkinsci/copyartifact-plugin/pull/81 are merged and I will rebase.

Thanks

ikedam:
#81 caused conflicts but merged manually.
Thanks!

ikedam:
Oops, I had to fix javadocs for changes of parameters in #81.
That caused Jenkins fail after merging.
Let me have time to fix that.

PR #79, opened 2016-03-07
[JENKINS-32526] - Handle null @AncestorInPath when selecting Jobs/Dow…

Dohbedoh:
…nstream/Upstream Jobs and for autocompletion.

[JENKINS-32526](https://issues.jenkins-ci.org/browse/JENKINS-32526)

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -577,9 +577,9 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
     public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
 
         public FormValidation doCheckProjectName(
-                @AncestorInPath AbstractProject anc, @QueryParameter String value) {
+                @AncestorInPath Job<?,?> anc, @QueryParameter String value) {
             // JENKINS-32526: Check that it behaves gracefully for an unknown context
-            if (anc == null) return FormValidation.ok();
+            if (anc == null) return FormValidation.ok(Messages.CopyArtifact_AncestorIsNull());
             // Require CONFIGURE permission on this project
             if (!anc.hasPermission(Item.CONFIGURE)) return FormValidation.ok();
             
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -577,7 +577,9 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
     public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
 
         public FormValidation doCheckProjectName(
-                @AncestorInPath Job<?,?> anc, @QueryParameter String value) {
+                @AncestorInPath AbstractProject anc, @QueryParameter String value) {
+            // JENKINS-32526: Check that it behaves gracefully for an unknown context
+            if (anc == null) return FormValidation.ok();
             // Require CONFIGURE permission on this project
             if (!anc.hasPermission(Item.CONFIGURE)) return FormValidation.ok();
             
--- a/src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
+++ b/src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
@@ -188,6 +188,11 @@ public class DownstreamBuildSelector extends BuildSelector {
             if (containsVariable(upstreamProjectName)) {
                 return FormValidation.ok();
             }
+
+            if (project == null) {
+                // Context is unknown and validation is useless.
+                return FormValidation.ok();
+            }
             
             Jenkins jenkins = Jenkins.getInstance();
             if (jenkins == null) {
@@ -239,7 +244,7 @@ public class DownstreamBuildSelector extends BuildSelector {
             }
             
             AbstractProject<?,?> upstreamProject = jenkins.getItem(
-                    upstreamProjectName, project.getRootProject(), AbstractProject.class
+                    upstreamProjectName, project == null ? null : project.getRootProject(), AbstractProject.class
             );
             if (upstreamProject == null || !upstreamProject.hasPermission(Item.READ)) {
                 return FormValidation.ok();
@@ -292,7 +297,9 @@ public class DownstreamBuildSelector extends BuildSelector {
                 @AncestorInPath AbstractProject<?,?> project
         ) {
             // Specified Item to allow to autocomplete folders (maybe confusing...).
-            return AutoCompletionCandidates.ofJobNames(Item.class, value, project, project.getParent());
+            return project == null
+                    ? AutoCompletionCandidates.ofJobNames(Item.class, value, null)
+                    : AutoCompletionCandidates.ofJobNames(Item.class, value, project, project.getParent());
         }
     }
 }
--- a/src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
+++ b/src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
@@ -194,10 +194,20 @@ public class DownstreamBuildSelector extends BuildSelector {
                 // Jenkins is unavailable and validation is useless.
                 return FormValidation.ok();
             }
-            
-            AbstractProject<?,?> upstreamProject = jenkins.getItem(
-                    upstreamProjectName, project, AbstractProject.class
+
+            if (project == null) {
+                // Context is unknown and validation is useless.
+                return FormValidation.ok(Messages.CopyArtifact_AncestorIsNull());
+            }
+
+            Job<?,?> upstreamRoot = (project instanceof AbstractProject)
+                    ? ((AbstractProject<?,?>) project).getRootProject()
+                    : project;
+
+            Job<?,?> upstreamProject = jenkins.getItem(
+                    upstreamProjectName, upstreamRoot, Job.class
             );
+
             if (upstreamProject == null || !upstreamProject.hasPermission(Item.READ)) {
                 return FormValidation.error(Messages.DownstreamBuildSelector_UpstreamProjectName_NotFound());
             }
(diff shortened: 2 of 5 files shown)


Dohbedoh:
@reviewbybees

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -577,7 +577,9 @@ public String resolve(String name) {
     public static final class DescriptorImpl extends BuildStepDescriptor<Builder> {
 
         public FormValidation doCheckProjectName(
-                @AncestorInPath Job<?,?> anc, @QueryParameter String value) {
+                @AncestorInPath AbstractProject anc, @QueryParameter String value) {
```
ikedam: Why do you need have it `AbstractProject`?
Pipeline (aka. workflow)'s snippet generator requires `Job`.
f28c55475678492a5e0659cb444b062fa5fe748b

src/test/java/hudson/plugins/copyartifact/DownstreamBuildSelectorTest.java
```diff
@@ -31,19 +31,10 @@
...
-import hudson.model.Cause;
-import hudson.model.FreeStyleProject;
-import hudson.model.FreeStyleBuild;
-import hudson.model.Item;
-import hudson.model.ParametersAction;
-import hudson.model.ParametersDefinitionProperty;
-import hudson.model.StringParameterDefinition;
-import hudson.model.StringParameterValue;
-import hudson.model.User;
-import hudson.model.Result;
```
ikedam: I don't like wildcard import. (especially for classes)
Do you actually need this change?

src/test/java/hudson/plugins/copyartifact/DownstreamBuildSelectorTest.java
```diff
@@ -605,14 +614,75 @@ public void testCheckUpstreamBuildNumber() throws Exception {
...
+
+    private void testAutoCompleteUpstreamProjectName(
+            String [] expectedValues,
+            String value,
+            AbstractProject project,
+            DownstreamBuildSelector.DescriptorImpl d) {
+
+        assertArrayEquals(expectedValues, d.doAutoCompleteUpstreamProjectName(value, project).getValues().toArray());
+        //JENKINS-32526
+        assertArrayEquals(expectedValues, d.doAutoCompleteUpstreamProjectName(value, null).getValues().toArray());
```
ikedam: It's better to sort the list or wrap it with `Set`.
`AutoCompleteCandidates` doesn't look ensure the order of values (strange behavior!)

src/test/java/hudson/plugins/copyartifact/DownstreamBuildSelectorTest.java
Dohbedoh: No sorry. That's the IDE that organized import automatically.

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -188,6 +188,11 @@ public FormValidation doCheckUpstreamProjectName(
             if (containsVariable(upstreamProjectName)) {
                 return FormValidation.ok();
             }
+
+            if (project == null) {
```
jglick: While you are it, you should replace `AbstractProject` with `Job` to get Pipeline compatibility missed in #55. Ditto in other related validation methods. CC @apemberton: currently with Pipeline this throws a `NullPointerException`; fixing the PR as written would make it unconditionally return OK; also fixing the type of the parameter would make it actually perform the desired validation.

jglick: [JENKINS-33578](https://issues.jenkins-ci.org/browse/JENKINS-33578) specifically.

Dohbedoh:
@jglick Changed the variable type to `Job`. I needed to workaround a call to `AbstractProject#getRootProject()` in the `doCheckUpstreamProjectName` and `doCheckUpstreamBuildNumber` functions.

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -194,9 +194,16 @@ public FormValidation doCheckUpstreamProjectName(
                 // Jenkins is unavailable and validation is useless.
                 return FormValidation.ok();
             }
+
+            AbstractProject<?,?> upstreamRoot;
+            if(project != null && project instanceof AbstractProject) {
+                upstreamRoot = ((AbstractProject) project).getRootProject();
+            } else {
+                upstreamRoot = null;
```
jglick: No, you want to set it to `project`.

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -194,9 +194,16 @@ public FormValidation doCheckUpstreamProjectName(
...
+            AbstractProject<?,?> upstreamRoot;
+            if(project != null && project instanceof AbstractProject) {
+                upstreamRoot = ((AbstractProject) project).getRootProject();
+            } else {
+                upstreamRoot = null;
+            }
             
             AbstractProject<?,?> upstreamProject = jenkins.getItem(
-                    upstreamProjectName, project.getRootProject(), AbstractProject.class
+                    upstreamProjectName, upstreamRoot, AbstractProject.class
```
jglick: And here you want to deal with `Job`, not `AbstractProject`.

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -237,9 +244,16 @@ public FormValidation doCheckUpstreamBuildNumber(
                 // Jenkins is unavailable and validation is useless.
                 return FormValidation.ok();
             }
+
+            AbstractProject<?,?> rootProject;
+            if(project != null && project instanceof AbstractProject) {
+                rootProject = ((AbstractProject) project).getRootProject();
+            } else {
+                rootProject = null;
```
jglick: ditto

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -237,9 +244,16 @@ public FormValidation doCheckUpstreamBuildNumber(
...
+            AbstractProject<?,?> rootProject;
+            if(project != null && project instanceof AbstractProject) {
+                rootProject = ((AbstractProject) project).getRootProject();
+            } else {
+                rootProject = null;
+            }
             
             AbstractProject<?,?> upstreamProject = jenkins.getItem(
-                    upstreamProjectName, project.getRootProject(), AbstractProject.class
+                    upstreamProjectName, rootProject, AbstractProject.class
```
jglick: ditto

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -292,7 +306,9 @@ public AutoCompletionCandidates doAutoCompleteUpstreamProjectName(
                 @AncestorInPath AbstractProject<?,?> project
```
jglick: Need to make this `Job`.

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -292,7 +306,9 @@ public AutoCompletionCandidates doAutoCompleteUpstreamProjectName(
                 @AncestorInPath AbstractProject<?,?> project
         ) {
             // Specified Item to allow to autocomplete folders (maybe confusing...).
-            return AutoCompletionCandidates.ofJobNames(Item.class, value, project, project.getParent());
+            return project == null
+                    ? AutoCompletionCandidates.ofJobNames(Item.class, value, null)
```
jglick: If there is no context, just turn off completions (return `new AutoCompletionCandidates()`).

jglick:
BTW if you want to _test_ Pipeline compatibility of completions, `WorkflowJob` is already in the test classpath. As I noted to @apemberton in [JENKINS-33577](https://issues.jenkins-ci.org/browse/JENKINS-33577), it is not really useful in this case anyway: `DownstreamBuildSelector` will not actually work.

src/main/resources/hudson/plugins/copyartifact/Messages.properties
```diff
@@ -1,3 +1,4 @@
+CopyArtifact.AncestorIsNull=Ancestor/Context Unknown: the project specified cannot be validated
```
ikedam: I don't think this is helpful for users.
How about a text like "You look configuring non-buildable job (may be template) ..."?

jglick: Well it is not exactly “non-buildable”. But certainly the term “ancestor” will be meaningless to a user (this is an API term). I think it suffices to say that the configuration form is being displayed without enough context to determine which job will host the build step.

ikedam:
You'd better add a comment when you add a new commit as github doesn't send a notification for commits.

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -196,7 +196,7 @@ public FormValidation doCheckUpstreamProjectName(
             }
             
             AbstractProject<?,?> upstreamProject = jenkins.getItem(
-                    upstreamProjectName, project.getRootProject(), AbstractProject.class
+                    upstreamProjectName, project, AbstractProject.class
```
ikedam: You might misunderstand the comment by jglick (https://github.com/jenkinsci/copyartifact-plugin/pull/79/files/65b30d6da1983e11be0f93d9defb664796fca99d#r56505288)
That meant you don't need to use `null` for `project` even when `project` doesn't have `getRootProject`.
`getRootProject` should be still required if it's available.

Though I don't know an actual case resulting `project.getRootProject() != project` in this context, it's better to implement in the same way as `DownstreamBuildSelector` works.

src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -177,7 +177,7 @@ protected boolean containsVariable(String str) {
          * @return
          */
         public FormValidation doCheckUpstreamProjectName(
-                @AncestorInPath AbstractProject<?,?> project,
+                @AncestorInPath Job<?,?> project,
```
ikedam: You should skip the validation if `project` is `null` (that is, when configuring a template).

Dohbedoh:
@ikedam You're right I misunderstood jglick's point. I reimplemented the use of `getRootProject` and skip the validation in case `project` is `null`.
@reviewbybees

ikedam:
Thanks for the update.
Looks good to me.

PR #80, opened 2016-03-16
[JENKINS-33577] Demonstration test

jglick:
[JENKINS-33577](https://issues.jenkins-ci.org/browse/JENKINS-33577)

@reviewbybees esp. @apemberton


pom.xml
```diff
@@ -22,7 +22,7 @@
     </licenses>
 
     <properties>
-        <workflow.version>1.1</workflow.version>
+        <workflow.version>1.4.2</workflow.version>
```
jglick: `build` did not return an appropriate value in the ancient 1.1 release. 1.4.2 is the newest compatible with 1.580.x.

pom.xml
```diff
@@ -24,7 +24,7 @@
     <properties>
       <jenkins.version>1.580.1</jenkins.version>
       <java.level>6</java.level>
-      <workflow.version>1.1</workflow.version>
+      <workflow.version>1.4.2</workflow.version>
```
jglick: `build` did not return an appropriate value in the ancient 1.1 release. 1.4.2 is the newest compatible with 1.580.x.

jglick:
@ikedam, @apemberton ping

ikedam:
@jglick 
The test looks good.
It demonstrates that `SpecificBuildSelector` and `build(p)` provides a feature like `DownstreamBuildSelector`.

But I don't think this change needs to be merged as this doesn't provide any fixes or new features of copyartifact, but contains changes for dependencies that isn't important for copyartifact itself.

Do you mean upgrading copyartifact-plugin should force also to upgrade pipeline-plugin?

jglick:
Well [1.4.2 is a rather old version](https://github.com/jenkinsci/pipeline-plugin/blob/master/CHANGES.md#142-jul-21-2015). Anyway this is only `test` scope so it has no effect on users.

ikedam:
> Anyway this is only test scope so it has no effect on users.

You're absolutely right.
Sorry, I misunderstood that copyartifact has an optional dependency to pipeline.
I found core provides everything necessary to support pipeline.

I think this is ready to merge.
Though this looks a test for pipiline's `build` method rather than a test for copyartifact, I think it makes sense also as a test to use `selector` with pipilene.

PR #84, opened 2016-04-03
[JENKINS-33578] DownstreamBuildSelector is not applicable to Workflow jobs.

ikedam:
[JENKINS-33578](https://issues.jenkins-ci.org/browse/JENKINS-33578)

Though the issue described in the ticket is fixed in #79, I found `DownstreamBuildSelector` is not applicable to Workflow jobs as it depends on [`AbstractBuild#getUpstreamRelationshipBuild`](http://javadoc.jenkins-ci.org/hudson/model/AbstractBuild.html#getUpstreamRelationshipBuild%28hudson.model.AbstractProject%29).

This change show error or warnings messages to users when they tries to use `DownstreamBuildSelector` with `Workflow`.

downstream_with_workflow


src/main/java/hudson/plugins/copyartifact/DownstreamBuildSelector.java
```diff
@@ -90,6 +90,7 @@ public String getUpstreamBuildNumber() {
     @Override
     protected boolean isSelectable(Run<?, ?> run, EnvVars env) {
         if (!(run instanceof AbstractBuild<?,?>)) {
+            // As this feature depends on `AbstractBuild#getUpstreamRelationshipBuild(AbstractProject<?,?>)`
             LOGGER.warning(String.format("Only applicable to AbstractBuild: but is %s.", run.getClass().getName()));
```
ikedam: This can never be warned at the configuration time...

PR #90, opened 2017-09-10
Fix test failures on ci.jenkins.io

ikedam:


src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
```diff
@@ -1367,7 +1367,8 @@ public void testWrappedCopierProjectNameSplit() throws Exception {
         rule.assertBuildStatusSuccess(b);
         FilePath fileToTest = b.getWorkspace().child("from-plain/tag.txt");
         assertTrue(fileToTest.exists());
-        assertEquals("jenkins-plain-2\n", fileToTest.readToString());
+        // Ignore line endings as it may differ for OS and autocrlf configurations.
+        assertEquals("jenkins-plain-2", StringUtils.trim(fileToTest.readToString()));
```
ikedam: https://ci.jenkins.io/job/Plugins/job/copyartifact-plugin/job/master/1/testReport/junit/hudson.plugins.copyartifact/CopyArtifactTest/testWrappedCopierProjectNameSplit/
```
expected:<jenkins-plain-2[]
> but was:<jenkins-plain-2[
]
>
```

I believe this is caused for the autocrlf configuration on the windows node of ci.jenkins.io.

ikedam:
Hmm... somehow `getDescriptorByType` in `Plugin#postInitialize` returns `null` on windows on ci.jenkins.io.
I'll try `InitMilestone` instead.

ikedam:
Thoungh I couldn't find the root cause of NPE, this change itself should be harmless and ready to merge.

PR #91, opened 2017-09-17
DO NOT MERGE: Add logs for detect when initializers called

ikedam:
To detect the root cause of #90.


ikedam:
Unfortunately it looks that I cannot get output logs from tests...

PR #88, opened 2017-09-07
[JENKINS-46700] symbol support for better pipeline compatibility

kohsuke:
See [the ticket](https://issues.jenkins-ci.org/browse/JENKINS-46700) for more details.

In order to allow symbol names on `BuildSelector` I had to turn them into fully-fledged class, since `@Symbol` cannot be placed on a field given the way it works.

PR #92, opened 2017-09-18
[JENKINS-46700] Add tests for symbols

ikedam:
[JENKINS-46700](https://issues.jenkins-ci.org/browse/JENKINS-46700)

Tests for #88 (this request includes commits in that request).

Using `Symbol` annotation in pipeline requires workflow-cps 2.10 and Jenkins >= 1.642.3.
Though I could having copyartifact continue to target Jenkins-1.580.1 as in the way of [run-selector-plugin](https://github.com/jenkinsci/run-selector-plugin/blob/run-selector-1.0.0/src/test/java/org/jenkinsci/plugins/runselector/steps/SelectRunStepTest.java#L164), it should cost to maintenance, and I decided to upgrade targeted Jenkins.


pom.xml
```diff
@@ -53,6 +53,11 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+          <groupId>org.jenkins-ci.plugins</groupId>
+          <artifactId>structs</artifactId>
+          <version>1.3</version>
```
ikedam: Changed to 1.3 for testing.
(1.2 causes `java.lang.NoClassDefFoundError: org/jenkinsci/plugins/structs/describable/UninstantiatedDescribable` in tests)

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -146,4 +156,173 @@ private WorkflowJob createWorkflow(String name, String script) throws IOExceptio
         job.setDefinition(new CpsFlowDefinition("node {" + script + "}", true));
         return job;
     }
+
+    // Tests for symbols.
+    // Test only that symbols are accessible as expected.
+    // No tests for features here as they should be tested in other tests.
+    // The latest pipeline supports calling `copyArtifacts` without `step`.
```
ikedam: Please let me know which version introduced the feature to call `SimpleBuildStep`s without `step()`
(I want to update this code comment).

ikedam:
I believe the test failure is caused for an issue of pipeline-library (cause everything works fine in my environment), and I'll retrigger the test later.

ikedam:
Tests passed.
@kohsuke @jglick Would you review this change?

pom.xml
jglick: Use a newer parent POM and you will see such mistakes in the `validate` phase, with a more meaningful error message.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
jglick: Oh, some rather old versions of various plugins. Too old to bother calling out at this point.

pom.xml
```diff
@@ -92,25 +97,37 @@
...
         <artifactId>workflow-job</artifactId>
-        <version>${workflow.version}</version>
+        <version>2.3</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-basic-steps</artifactId>
-        <version>${workflow.version}</version>
+        <version>1.15</version>
```
jglick: Use a 2.x version.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -26,34 +26,44 @@
...
 
 /**
  * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
  */
 public class CopyArtifactWorkflowTest {
 
-    @Rule
-    public JenkinsRule jenkinsRule = new JenkinsRule();
+    @ClassRule
+    public static JenkinsRule jenkinsRule = new JenkinsRule();
```
jglick: Be careful. Yes you can speed up test execution, but you can also introduce weird inter-test dependencies that are tricky to debug.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -146,4 +156,173 @@ private WorkflowJob createWorkflow(String name, String script) throws IOExceptio
...
+        copiee.getBuildersList().add(ca);
+        copiee.getBuildersList().add(new FileWriteBuilder("artifact.txt", "foobar"));
+        copiee.getPublishersList().add(new ArtifactArchiver("artifact.txt"));
+
+        jenkinsRule.assertBuildStatusSuccess(upstream.scheduleBuild2(0));
+        jenkinsRule.assertBuildStatusSuccess(copiee.scheduleBuild2(0));
+
+        WorkflowJob copier = createWorkflow(
+            "copier",
+            "step(copyArtifacts(projectName: 'copiee', selector: downstream(upstreamProjectName: 'upstream', upstreamBuildNumber: '1')));"
```
jglick: Oh I see what you are saying—you did not in fact get the test to work properly without `step`. Meaning you need to update, IIRC, `workflow-basic-steps`.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -146,4 +156,173 @@ private WorkflowJob createWorkflow(String name, String script) throws IOExceptio
...
+        WorkflowJob copiee = createWorkflow(
+            "copiee",
+            "writeFile text: 'foobar', file: 'artifact.txt';"
+            + "archive includes: 'artifact.txt';"
+        );
+        jenkinsRule.assertBuildStatusSuccess(copiee.scheduleBuild2(0));
+
+        WorkflowJob copier = createWorkflow(
+            "copier",
+            "step(copyArtifacts(projectName: 'copiee', selector: buildParameter('<StatusBuildSelector/>')));"
```
jglick: What?! Why is this taking an XML argument?

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -146,4 +156,173 @@ private WorkflowJob createWorkflow(String name, String script) throws IOExceptio
...
+            + "echo readFile('artifact.txt');"
+        );
+        jenkinsRule.assertLogContains("foobar", jenkinsRule.assertBuildStatusSuccess(copier.scheduleBuild2(0)));
+    }
+
+    @Test
+    public void testSpecificBuildSelector() throws Exception {
+        WorkflowJob copier = createWorkflow(
+            "copier",
+            "step(copyArtifacts(projectName: 'copiee', selector: specific(\"${build('copiee').number}\")));"
```
jglick: or

```groovy
selector: specific(build('copiee').id)
```

(BTW `SpecificBuildSelector` should perhaps use `Job.getBuild(String)` rather than looking for arguments that seem to be numbers.)

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -146,4 +156,173 @@ private WorkflowJob createWorkflow(String name, String script) throws IOExceptio
...
+            "step(copyArtifacts(projectName: 'copiee', selector: upstream()));"
+            + "echo readFile('artifact.txt');"
+        );
+        jenkinsRule.assertBuildStatusSuccess(copiee.scheduleBuild2(0));
+        jenkinsRule.assertLogContains("foobar", jenkinsRule.assertBuildStatusSuccess(copier.getLastBuild()));
+    }
+
+    @Test
+    public void testWorkspaceBuildSelector() throws Exception {
+        // workspace is useless for pipeline jobs
```
jglick: Then delete the test.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
ikedam: I used versions used in workflow-cps 2.10.
OK, I'll use the latest versions available for Jenkins-1.642.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
ikedam: `ParameterizedBuildSelector` takes a parameter created by `BuildSelectorParameter` which is a Xstream expression of a selector.
But this test is confusing and I'll update it to recieve a build parameter.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
ikedam: I found this don't speed up the test execution much. (120secs vs 130 secs for `CopyArtifactWorkflowTest`)
Reverted this change.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
ikedam: Created https://issues.jenkins-ci.org/browse/JENKINS-47074

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
ikedam: I actually believe `workspace` feature is useless for pipelines, but want to keep this test to ensure the symbol works.
Changed the comment to clarify the limitation of `workspace` (that is, only applicable to `AbstractProject`s).

src/test/resources/hudson/plugins/copyartifact/CopyArtifactTest/testPermissionWhenParameterized/config.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
ikedam: Extracted from CopyArtifactTest.zip to avoid https://issues.jenkins-ci.org/browse/JENKINS-47075

ikedam:
Updated the request:
* Use the latest plugin-pom and pipeline plugins in tests
* Expanded CopyArtifactTest.zip to avoid [JENKINS-47075](https://issues.jenkins-ci.org/browse/JENKINS-47075)

ikedam:
@jglick I addressed your comments.
Would you review this request again?

PR #97, opened 2017-12-06
POM update

jglick:
@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.35</version>
+        <version>2.37</version>
+        <relativePath />
     </parent>
 
     <artifactId>copyartifact</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -4,20 +4,21 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.35</version>
+        <version>2.37</version>
```
jglick: Possible to jump to 3.x while still on a Jenkins 1.x baseline, but officially only Jenkins 2.x is supported.

src/test/java/hudson/plugins/copyartifact/CopyArtifactTest.java
```diff
@@ -1861,7 +1861,7 @@ public void testSymlinksInDirectory() throws Exception {
     }
     
     private static class TestQueueItemAuthenticator extends jenkins.security.QueueItemAuthenticator {
-        private final org.acegisecurity.Authentication auth;
+        private final transient org.acegisecurity.Authentication auth;
```
jglick: cf. https://github.com/jenkinsci/jenkins/pull/3120

jglick:
@reviewbybees done

PR #98, opened 2018-02-20
[JENKINS-18662] Strengthen test to differentiate between explicit and default excludes

jglick:
Amends test from #42. I found that the tests which supposedly are exercising explicit excludes (JENKINS-18662) in fact pass if Ant’s default excludes are applied (contrary to JENKINS-14900) but explicit excludes are totally ignored. This is because the choice of explicit excludes happened to be files which default excludes also cover. This was confusing and not helpful for verifying the actual behavior.

Also the tests did not check the use of _multiple_ excludes.

@reviewbybees


jglick:
@reviewbybees done

PR #71, opened 2015-09-23
[WIP] [JENKINS-24892] Copyartifact 2.0

ikedam:
[JENKINS-24892](https://issues.jenkins-ci.org/browse/JENKINS-24892)

Refactoring of copyartifact.

New features:
- Build-filter based configuration (https://issues.jenkins-ci.org/browse/JENKINS-24890)
- You can select a "Copy Operation":
  - Copy standard artifact files: copy artifacts from `StandardArtifactManager` (works as before)
  - Copy extended artifact files: copy arftifacts from generic `ArtifactManager` (https://issues.jenkins-ci.org/browse/JENKINS-22637)
    - It doesn't preserver file modes and symbolic links (as `VirtualFile` doesn't provide that feature)
  - Copy files from workspace: works as `WorkspaceSelector` did. (`WorkspaceSelector` is no longer supported)

(Incompatible) changes:
- [`Copier`](https://github.com/jenkinsci/copyartifact-plugin/blob/master/src/main/java/hudson/plugins/copyartifact/Copier.java) is removed.
  - You can extend `CopyArtifactOperation` instead.
- No longer supports `WorkspaceSelector` (https://github.com/jenkinsci/copyartifact-plugin/blob/master/src/main/java/hudson/plugins/copyartifact/WorkspaceSelector.java).
  - You can use `CopyWorkspaceFiles` for `CopyOperation` instead.
- Permission checks are performed only at runtime (https://issues.jenkins-ci.org/browse/JENKINS-24888)

Known issues:
- `VirtualFile` (`ArtifactManager`) doesn't support listing files with ant file patttens (like `**/*.java`) and I had to reimplement it by myself.
  - There can be bugs.
  - There can be performance issues.

This request is still in progress. Left tasks are:
- ~~Make tests passed.~~
- ~~UI: remove `parameters` and handle `BuildFilter`.~~
- ~~Add copy operation abstraction.~~
- ~~Add more `BuildFilter`s including `AndBuildFilter` and `OrBuildFilter`.~~
- ~~Add fallback build selector.~~
- ~~Migrate existing `BuildSelector`s to `BuildFilter`s and support automatic configuration upgrade.~~
- ~~Extend `ParameterizedBuildSelector` to support filters.~~
- ~~Add texts.~~
- Add new tests.

PR #99, opened 2018-02-20
[JENKINS-22637] ArtifactManager support

jglick:
[JENKINS-22637](https://issues.jenkins-ci.org/browse/JENKINS-22637): supporting custom `ArtifactManager`s via `VirtualFile`. Unlike #71 the goal here is to make just this change with as little impact on existing installations as possible.

- [X] basic refactoring
- [x] integration test with the `compress-artifacts` plugin (14 votes)
- [X] restore support for explicit excludes (1 vote)
- [ ] restore support for symlinks (2 votes): #100
- [ ] restore support for file attribute handling (11 votes): #100
- [ ] restore support for default excludes (1 vote): #100

To handle symlinks and file attributes may require new core APIs—TBD. Really if you care about platform-specific stuff like this you should create a tarball and archive that instead, since you cannot guarantee that the Jenkins master will preserve these nuances. Similarly, suppressing of default excludes is desirable but might just be treated as a core bug fix, and again it is going to be safer and much more efficient to create a ZIP or tarball of all your little files.

I also plan to

- [ ] support direct download as per https://github.com/jenkinsci/jenkins/pull/3302: #100

but this will certainly require a new core baseline and might better be filed as an add-on PR.

Note that I am dropping the `Copier` extension point here. In practice I doubt this would be incompatible for anyone:
* There are [no known external implementations](https://jenkins.io/doc/developer/extensions/copyartifact/).
* This never really behaved like a regular extension point to begin with: the user cannot select an implementation! In fact `FilePathCopyMethod` was dead code—even as of 3cf80e567c8171428be377ddb2ec6855a7570dc3, it was always overridden by `FingerprintingCopyMethod`. And that one was rather misnamed as of #11—it may or may not do fingerprinting!
* [JENKINS-7753](https://issues.jenkins-ci.org/browse/JENKINS-7753) claimed as justification
  > I would like to implement a plug-in which provides CopyArtifact with rsync capabilities when copying workspaces between jobs. This will decrease load on the master and most likely increase IO-throughput.
  
  which is sensible enough but better handled via `ArtifactManager`, especially as of [JENKINS-49635](https://issues.jenkins-ci.org/browse/JENKINS-49635).

I am also making an API change to `BuildSelector`. In this case there [_are_ two known external implementations](https://jenkins.io/doc/developer/extensions/copyartifact/) but neither override the `getSourceDirectory` method. In fact, other than `WorkspaceSelector`, it is hard to imagine any other useful override. I am making a cursory effort to retain compatibility with any external overrides just in case.

Subsumes #98.

@reviewbybees


src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -412,7 +426,7 @@ public void perform(@Nonnull Run<?, ?> build, @Nonnull FilePath workspace, @Nonn
                 throw new AbortException(message);
             }
         }
-        FilePath targetDir = workspace, baseTargetDir = targetDir;
```
ikedam: Note for me: `baseTargerDir` looks not used at all. Completely useless.

src/main/java/hudson/plugins/copyartifact/Copier.java
```diff
@@ -21,7 +21,9 @@
  * @author Alan Harder
  * @author Kohsuke Kawaguchi
  * @see "JENKINS-7753"
+ * @deprecated No longer used.
  */
+@Deprecated
 public abstract class Copier implements ExtensionPoint {
```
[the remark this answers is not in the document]
jglick: Well, no “exception” would result from using them somehow. I do not have a strong opinion either way.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: Right, AFAICT it was never used anyway.

jglick:
@ikedam so I am leaving it up to your judgment what to do here. I noted the number of JIRA votes for various affected issues, to the extent that is meaningful data. This PR in isolation implements a feature with 14 votes (and no good workaround), and regresses three bug reports with a combined total of 14 votes (and the workaround of archiving a tarball rather than individual files). #100 implements the feature (and more) without regressing any of the prior bugs, but will require a new core baseline. :man_shrugging:

One option not previously discussed would be to merge this but release only as a `-beta` version, allowing users interested in JENKINS-22637 to get the implementation from the experimental update center; and then release #100 if and when the corresponding core PR is released (the update center is smart enough not to offer that update to anyone using older LTS lines).

The hope is that once JEP-301 is deployed, Essentials users would get new stuff that just works, right away. Obviously a lot needs to happen before that is a reality.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -512,33 +523,112 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
-            int cnt;
-            if (!isFlatten())
-                cnt = copier.copyAll(srcDir, expandedFilter, expandedExcludes, targetDir, isFingerprintArtifacts());
-            else {
-                targetDir.mkdirs();  // Create target if needed
-                FilePath[] list = srcDir.list(expandedFilter, expandedExcludes, false);
-                for (FilePath file : list)
-                    copier.copyOne(file, new FilePath(targetDir, file.getName()), isFingerprintArtifacts());
-                cnt = list.length;
+            targetDir.mkdirs();  // Create target if needed
```
oleg-nenashev: would be nice to switch it to Java7 Files API while you are around

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -512,33 +523,112 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
 
-        copier.initialize(src, dst, srcDir, baseTargetDir);
+        Map<String, String> fingerprints;
+        MessageDigest md5;
+        if (isFingerprintArtifacts()) {
+            fingerprints = new HashMap<>();
+            try {
+                md5 = MessageDigest.getInstance("MD5");
+            } catch (NoSuchAlgorithmException x) {
+                throw new AssertionError(x);
```
oleg-nenashev: does not really matter, but maybe just IOException? There were some discussions about disabling MD5 at all to harden the security. I'd guess it is not realistic, but I would rather prefer to avoid errors in the production code

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -512,33 +523,112 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
+                        } else {
+                            r.addAction(new Fingerprinter.FingerprintAction(r, fingerprints));
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private static List<String> scan(VirtualFile root, String expandedFilter, @CheckForNull String expandedExcludes) throws IOException {
```
oleg-nenashev: Maybe it should be moved to core utils eventually. Looks to be generic enough

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -512,33 +523,112 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
-            else {
-                targetDir.mkdirs();  // Create target if needed
-                FilePath[] list = srcDir.list(expandedFilter, expandedExcludes, false);
-                for (FilePath file : list)
-                    copier.copyOne(file, new FilePath(targetDir, file.getName()), isFingerprintArtifacts());
-                cnt = list.length;
+            targetDir.mkdirs();  // Create target if needed
+            List<String> list = scan(srcDir, expandedFilter, expandedExcludes);
+            for (String file : list) {
+                copyOne(src, dst, fingerprints, srcDir.child(file), new FilePath(targetDir, isFlatten() ? file.replaceFirst(".+/", "") : file), md5);
```
oleg-nenashev: passing `isFingerprintArtifacts()` looked better, but OK

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -512,33 +523,112 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
+            int cnt = list.size();
             console.println(Messages.CopyArtifact_Copied(cnt, HyperlinkNote.encodeTo('/'+ src.getParent().getUrl(), src.getParent().getFullDisplayName()),
                     HyperlinkNote.encodeTo('/'+src.getUrl(), Integer.toString(src.getNumber()))));
             // Fail build if 0 files copied unless copy is optional
             return cnt > 0 || isOptional();
         } finally {
-            copier.end();
+            if (fingerprints != null) {
+                for (Run<?, ?> r : new Run<?, ?>[] {src, dst}) {
+                    if (fingerprints.size() > 0) {
```
oleg-nenashev: It is not modified inside the cycle IIUC, maybe makes sense to just move to the outer check

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -512,33 +523,112 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
+    /** Similar to a method in {@link DirectoryScanner}. */
+    private static String normalizePattern(String p) {
+        String pattern = p.replace('\\', '/'); // we only deal with forward slashes here
+        if (pattern.endsWith("/")) {
+            pattern += SelectorUtils.DEEP_TREE_MATCH;
+        }
+        return pattern;
+    }
+
+    private static void copyOne(Run<?,?> src, Run<?,?> dst, Map<String, String> fingerprints, VirtualFile s, FilePath d, MessageDigest md5) throws IOException, InterruptedException {
```
oleg-nenashev: same, does not seem to be plugin-specific

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
ikedam: Note for me: snipping ".+/" is equivalent to the current implementation calling `FilePath#getName()` in `FingerprintingCopyMethod`.
(I'm not sure this is a good behavior as reported in [JENKINS-16987](https://issues.jenkins-ci.org/browse/JENKINS-16987))

ikedam:
@jglick Sorry for my late response.

* Releasing to the experimental update center at first
* Releasing to the actual update center when jenkinsci/jenkins#3302 is released

This option looks good to me.

Let me have a little more time to work for https://issues.jenkins-ci.org/browse/JENKINS-47905 and https://issues.jenkins-ci.org/browse/JENKINS-47074.
So releases:

* copyartifact-1.40: JENKINS-47905 , JENKINS-47074
* copyartifact-1.41-beta: JENKINS-22637
* copyartifact-1.41: JENKINS-22637

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: To do that we would first need to reimplement fingerprints in Jenkins core, which would surely force changes here as well anyway.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: But now we need to pass in the map.

jglick: No, this `replaceFirst` call just implements the flatten mode. It has nothing to do with fingerprinting. The behavior you mention is retained in `copyOne` as

```java
fingerprints.put(s.getName(), digest);
//                ^^^^^^^^^^
```

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: Does not really matter—the loop has only two iterations.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: This code is deleted in #100.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
jglick: It is pretty specific to `copyartifact` I think.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
ikedam: Oops, I misplaced the comment!
https://github.com/jenkinsci/copyartifact-plugin/pull/100/files#r174750602

ikedam:
As JENKINS-47074 will be resolved as Won't fix and JENKINS-47905 provides no new features, versions are:

copyartifact-1.39.1: JENKINS-47905
copyartifact-1.40-beta: JENKINS-22637
copyartifact-1.40: JENKINS-22637

ikedam:
Really sorry for long absence.
I resume this task.
Though the extension of `VirtualFile` is already released as Jenkins2.108, I believe it still make sense to release 1.40-beta for LTS users.

ikedam:
Released 1.40-beta-1 with this change.
I couldn't update the wiki page as login failure (looks like a system error). I'll try it later.

jglick:
> releasing 1.40 will mask 1.40-beta

AFAIK not, since the UC for LTS users should only be offering 1.39.

ikedam:
Updates center provides several versions of plugin lists for core versions by redirection.
* https://updates.jenkins.io/update-center.json?id=default&version=2.118 is redirected to https://updates.jenkins.io/current/update-center.json
* https://updates.jenkins.io/update-center.json?id=default&version=2.107.2 is redirected to https://updates.jenkins.io/2.107/update-center.json

But the experimental update center provides only one version, and will result 1.40-beta-1 inaccessible via the update center.

Anyway, that doesn't matter so much as users can download the beta version directly from http://updates.jenkins-ci.org/download/plugins/copyartifact/1.40-beta-1/copyartifact.hpi .

PR #100, opened 2018-02-21
[JENKINS-22637] ArtifactManager support on new core

jglick:
Subsumes #99 but extended to take advantage of https://github.com/jenkinsci/jenkins/pull/3302. Also downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/67.

@reviewbybees

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-alpha-3</version>
+    <version>1.0-beta-1</version>
   </extension>
 </extensions>
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <extension>
+    <groupId>io.jenkins.tools</groupId>
+    <artifactId>git-changelist-maven-extension</artifactId>
+    <version>1.0-alpha-3</version>
+  </extension>
+</extensions>
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-2</version>
+    <version>1.0-beta-3</version>
   </extension>
 </extensions>
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,7 +1,7 @@
 <extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
   <extension>
-    <groupId>io.jenkins.tools</groupId>
+    <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-1</version>
+    <version>1.0-beta-2</version>
   </extension>
 </extensions>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.10-20180509.203140-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/105 -->
+        <version>3.10</version>
         <relativePath />
     </parent>
 
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.7-20180326.223441-1</version> <!-- TODO https://github.com/jenkinsci/plugin-pom/pull/98 -->
+        <version>3.7</version>
         <relativePath />
     </parent>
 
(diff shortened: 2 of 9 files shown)


jglick:
Last build failure is I think due to a conflict between deployed snapshots. Even when specifying timestamps in the POM, Maven does not differentiate between different timestamps of the same `*-SNAPSHOT` version in the local repository when constructing the `compile` and other classpaths, as I have pointed out to @stephenc (but probably I need to file a MNG for this). Thus if the build machine happens to have downloaded multiple variants to the local repo, presumably because of some unrelated job using a different timestamp dependency, everything blows up. The best workaround would be to run every build in a new container/VM, or at least wrap the `mvn` build in a container, which AFAIK we do not currently do for regular plugin builds (@rtyler correct me if I am mistaken here).

jglick:
Alternately, we could just use a per-workspace local Maven repository in `buildPlugin.groovy`.

pom.xml
```diff
@@ -151,12 +160,18 @@
...
+        <version>1.10</version>
         <scope>test</scope>
       </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-api</artifactId>
+            <version>${workflow-api-plugin.version}</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
```
ikedam: inconsistent indentation.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -512,33 +522,131 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
                     HyperlinkNote.encodeTo('/'+src.getUrl(), Integer.toString(src.getNumber()))));
             // Fail build if 0 files copied unless copy is optional
             return cnt > 0 || isOptional();
         } finally {
-            copier.end();
+            if (fingerprints != null) {
+                Map<String, String> fingerprintsShallow = new HashMap<>();
+                FingerprintMap map = Jenkins.get().getFingerprintMap();
+                for (Map.Entry<String, String> entry : fingerprints.entrySet()) {
+                    String name = entry.getKey().replaceFirst(".+/", "");
```
ikedam: Note for me: snipping ".+/" is equivalent to the current implementation calling `FilePath#getName()` in `FingerprintingCopyMethod`.
(I'm not sure this is a good behavior as reported in [JENKINS-16987](https://issues.jenkins-ci.org/browse/JENKINS-16987))

pom.xml
```diff
@@ -35,7 +37,7 @@
         <artifactId>maven-plugin</artifactId>
         <version>2.7.1</version>
         <optional>true</optional>
-        <exclusions>
+        <exclusions> <!-- TODO use 3.0 -->
```
oleg-nenashev: Just upgrade it to 3.1 in a separate PR?

jglick: Maybe. Up to maintainer I guess.

src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
```diff
@@ -523,115 +524,132 @@ private static ItemGroup getItemGroup(Run<?, ?> build) {
...
         if (srcDir == null) {
             return isOptional();  // Fail build unless copy is optional
         }
-
-        Map<String, String> fingerprints;
-        MessageDigest md5;
-        if (isFingerprintArtifacts()) {
-            fingerprints = new HashMap<>();
-            try {
-                md5 = MessageDigest.getInstance("MD5");
```
jglick: Note that after the merge of #99, the diff of _this_ PR against `master` actually got longer, since it more closely followed the code in 1.39. (IIRC)

pom.xml
ikedam: I agree to use maven-plugin 3.0+ to simplify the dependency.
I'll do that before releasing the next version.

I prefer to use the least version and plan to use 3.0, but let me know if 3.1 (or 3.1.2, the latest one) is better than 3.0.

.mvn/extensions.xml
```diff
@@ -0,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <extension>
+    <groupId>io.jenkins.tools</groupId>
+    <artifactId>git-changelist-maven-extension</artifactId>
```
ikedam: Would you let me know how this extension work? There look no documentations about this extension.
I wonder the version rewriting in `release:prepare` and `release:perform` works correct.

ikedam:
@jglick 
Though this uses the beta version of workflow-api-plugin, it's used for only testing purpose and I believe it never cause problems in releasing the plugin. Is it right?
Let me know if I have to do something before merging this and releasing the new version
(I believe I can simply remove `on-hold` from the subject).

pom.xml
jglick: I would recommend the latest.

jglick:
> Though this uses the beta version of workflow-api-plugin, it's used for only testing purpose

True enough:

```manifest
Plugin-Dependencies: maven-plugin:2.7.1;resolution:=optional,matrix-pr
 oject:1.3,structs:1.14
```

Nor is it present in the “flattened” POM.

(Unlike, say, https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/60 or `artifact-manager-s3-plugin` current `master`, which need the `workflow-api` update at runtime.)

> it never cause problems in releasing the plugin

Not that I can think of. `maven-release-plugin` will refuse to proceed with a _snapshot_ dependency, even `test`-scoped, but this is not a snapshot. I do plan to introduce Enforcer rules to prevent you from accidentally releasing a plugin with an incremental dependency, but as noted in [JENKINS-50804](https://issues.jenkins-ci.org/browse/JENKINS-50804) this could probably just ignore `test` scope.

If https://github.com/jenkinsci/workflow-api-plugin/pull/67 is _rejected_ for some reason then of course we would need to figure something out in the long run, but there would be no immediate harm.

.mvn/extensions.xml
ikedam: Thanks! I got what incremental is.

So `useBeta=true` only for git-changelist-maven-extension and not for workflow-api.
I will remove `useBeta=true` when the actual version of  git-changelist-maven-extension is released.

pom.xml
ikedam: Ok, I'll update to the latest.

ikedam:
Ok, I'll try to make a new release (After upgrading the version of maven-plugin).
I know I have to update git-changelist-maven-extension and workflow-api when new versions of them are released.

PR #101, opened 2018-03-14
[FIXED JENKINS-47074] Support build ID in SpecificBuildSelector

ikedam:
https://issues.jenkins-ci.org/browse/JENKINS-47074

It makes it easier to use SpecificBuildSelector in pipeline jobs like this:
* Before
    ```
    copyArtifacts(projectName: "fromJob", selector: specific("${build("fromJob").number}")
    ```
* After
    ```
    copyArtifacts(projectName: "fromJob", selector: specific("build("fromJob").id")
    ```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -193,6 +193,19 @@
       <tag>HEAD</tag>
   </scm>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.jenkins-ci.tools</groupId>
+        <artifactId>maven-hpi-plugin</artifactId>
+        <configuration>
+          <!-- JENKINS-47074: Build ID precedes display names. -->
+          <compatibleSinceVersion>1.40</compatibleSinceVersion>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
   <repositories>
     <repository>
       <id>repo.jenkins-ci.org</id>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -193,6 +193,19 @@
...
+        <groupId>org.jenkins-ci.tools</groupId>
+        <artifactId>maven-hpi-plugin</artifactId>
+        <configuration>
+          <!-- JENKINS-47074: Build ID precedes display names. -->
+          <compatibleSinceVersion>1.40</compatibleSinceVersion>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
```
ikedam: This change may cause regressions if `yyyy-MM-dd_HH-mm-ss` is used for display names.
(I don't think it's often the case, though).

jglick:
Nonnumeric IDs do not exist after, IIRC, 1.597 or something, which is older than your baseline. So it would suffice to just accept `String` in place of `int`.

ikedam:
I didn’t know that.
So copyartifact already provides a sufficient feature.
I’ll add a test to see that `build.id` can replace `”${build.number}”`.

jglick:
(adding label based on last comment; adjust if I misunderstood)

ikedam:
I added a test to verify that `SpecificBuildSelector` accepts build IDs in #102 .
We don't need to make changes for [JENKINS-47074](https://issues.jenkins-ci.org/browse/JENKINS-47074), and I'll make a new release 1.39.1 only with #96 in the next week.

PR #104, opened 2018-06-05
Using RobustHTTPClient utility for artifact downloads

jglick:
Picks up https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin/pull/9 so that network problems need not be fatal. Follows up #100.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.13</version>
+        <version>3.15</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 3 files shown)


jglick:
> Will the annotation be removed

Not that quickly, no.

> it looks succeed even without `useBeta=true` for `apache-httpcomponents-client-4-api-plugin`

The property is set on the plugin _consuming_ the beta API.

ikedam:
> > it looks succeed even without useBeta=true for apache-httpcomponents-client-4-api-plugin
> The property is set on the plugin consuming the beta API.

Oh, I see. I misunderstood that.

It looks annotated as beta for `VirtualFile#toExternalURL` is beta.
https://github.com/jenkinsci/jenkins/blob/jenkins-2.121/core/src/main/java/jenkins/util/VirtualFile.java#L352

jglick:
@ikedam do you have a plan to release this? https://github.com/jenkinsci/jenkins/pull/3480 remains on hold until then.

ikedam:
@jglick
I plan to make a new release after resolving #105.

ikedam:
@jglick I released copyartifact-1.41 with this change.

PR #105, opened 2018-06-19
Add pipeline property copyArtifactPermission

JeremyMarshall:
Allows this in the options block

```
pipeline {
    agent any
    options {
        copyArtifactPermission('job1,...')
    }
    stages{...}
}
```
as discussed [here](https://groups.google.com/forum/#!topic/jenkinsci-dev/nJ54NaM7ctU)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>cloudbees-folder</artifactId>
-        <version>4.0</version>
+        <version>5.16</version>
         <scope>test</scope>
       </dependency>
       <dependency>
@@ -102,6 +102,7 @@
         <artifactId>workflow-cps</artifactId>
         <version>2.32</version>
         <scope>test</scope>
+        <classifier>tests</classifier>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -168,6 +169,11 @@
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>scm-api</artifactId>
         <version>2.0.8</version>
+      </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-multibranch</artifactId>
+        <version>2.10</version>
         <scope>test</scope>
       </dependency>
     </dependencies>
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifactPermissionProperty.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifactPermissionProperty.java
@@ -53,6 +53,7 @@ import hudson.model.JobProperty;
 import hudson.model.JobPropertyDescriptor;
 import hudson.model.AbstractProject;
 import hudson.util.FormValidation;
+import org.jenkinsci.Symbol;
 
 /**
  *　Job Property to define projects that can copy artifacts of this project.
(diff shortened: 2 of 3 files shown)


src/main/java/hudson/plugins/copyartifact/CopyArtifactPermissionProperty.java
```diff
@@ -53,10 +53,13 @@
 import hudson.model.JobPropertyDescriptor;
 import hudson.model.AbstractProject;
 import hudson.util.FormValidation;
+import org.jenkinsci.Symbol;
+import org.kohsuke.stapler.export.ExportedBean;
 
 /**
  *　Job Property to define projects that can copy artifacts of this project.
  */
+@ExportedBean
```
jglick: Does not seem to make sense, because it has no exported attributes AFAICT. For example, you could add to `getProjectNameList()`:

```java
@Exported(name = "projects")
```

Maybe better to file that as a separate PR with explanation, and keep this one for the `@Symbol` which should be unrelated?

JeremyMarshall: Remove. A bit of cargo culting I thought it helped with @Symbol...

JeremyMarshall:
@ikedam I added a test. Its not the most straightforward thing to do as adding properties into `CpsFlowDefinition` doesn't load properties/options. I ended up doing it like [matrix-auth-plugin](https://github.com/jenkinsci/matrix-auth-plugin/blob/a1e6e2171f90c5d5e9a502ff184dbc7ca0d7e54e/src/test/java/hudson/security/AuthorizationMatrixPropertyTest.java#L31)

jglick:
Someone needs to resolve the merge conflict.

pom.xml
```diff
@@ -113,6 +113,7 @@
         <artifactId>workflow-cps</artifactId>
         <version>2.32</version>
         <scope>test</scope>
+        <classifier>tests</classifier>
```
jglick: I think you need to keep the original dep, _and_ add the `tests`-scoped dep. To avoid then duplicating the `version`, normally this is extracted to a property like `workflow-cps-plugin.version`.

ikedam:
I recommend to use `git rebase` to resolve the conflict.
Merging master results diff in this request gets messy and makes it difficult to see what's the actual change in this request.

JeremyMarshall:
I didn't rebase as I was worried there might be other checkouts out in the will but will do so

I looked into the test tooling to try to resolve adding as an option/property and couldn't find any so I'll happily give that away as its a far larger piece.

There is another test failure due to the credentials plugin and how I bumped the versions

pom.xml
```diff
@@ -59,7 +59,7 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>cloudbees-folder</artifactId>
-        <version>4.0</version>
+        <version>5.16</version>
```
jglick:
```
Folders Plugin v5.16 failed to load.
 - Credentials Plugin v1.24 is older than required. To fix, install v2.1.0 or later.
```

Means you need to add a `<dependencyManagement>` section requesting `credentials` in (at least) `2.1.0`.

Normally these sorts of problems are caught during `mvn validate` when using a sufficiently new parent POM, but there are some edge cases such as use of `optional` dependencies (probably the case here) which break that. So to check effectiveness, use

    mvn test -Dtest=InjectedTest

pom.xml
jglick: (still true)

ikedam:
> > Merging master results diff in this request gets messy
> It should not. GitHub is smart and properly tracks what commits went where, so long as you avoid destructive history operations.

Sorry, so I should be a case for another repository management system.

ikedam:
> I have unrelated stuff whose release is waiting on this, so maybe I should file a subsuming PR to get it moving along?

No need.
I’ll release 1.40.1 this weekend if this request can’t be resolved.

jglick:
Check #106.

PR #106, opened 2018-06-27
copyArtifactPermission II

jglick:
Trying to fix up #105.

It took me a long time to get the new test to pass, and it turns out that this is because the conditions under which this feature are actually useful for Pipeline are so rarefied that it might as well not exist. It does not work if you are using a `QueueItemAuthenticator` like `authorize-project`—because if the build authentication lacks `READ` permission to upstream, [this line](https://github.com/jenkinsci/copyartifact-plugin/blob/3680ce8e6e2db2d19e66170d1cc44aa0c25d0b13/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java#L411) returns null even though the project name is correct. It does not work if you simply specify the upstream project name as a string literal, since then [this line](https://github.com/jenkinsci/copyartifact-plugin/blob/3680ce8e6e2db2d19e66170d1cc44aa0c25d0b13/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java#L412) will not match and `CopyArtifactPermissionProperty` is not even consulted. It does not work if you define a parameter for the upstream project name and then use Groovy interpolation to resolve that parameter value before passing it to the step, as is recommended for every other case where Pipeline scripts use parameters in steps. It does not work in the first build if you used the `properties` step to define the parameter, since you cannot use the `${params.UPSTREAM}` syntax via Groovy interpolation (as before), and [this line](https://github.com/jenkinsci/copyartifact-plugin/blob/3680ce8e6e2db2d19e66170d1cc44aa0c25d0b13/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java#L396) will not match anything since there were no parameters defined when the build was scheduled and thus no `ParametersAction`. It _only_ works if you specifically decline to set per-build authentication, in which case it is unclear why [this code](https://github.com/jenkinsci/copyartifact-plugin/blob/3680ce8e6e2db2d19e66170d1cc44aa0c25d0b13/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java#L490-L498) even exists, and if you predefined the parameter (not an option for multibranch projects), and if you used a freestyle-like parameter reference rather than the standard Pipeline idioms.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>cloudbees-folder</artifactId>
-        <version>4.0</version>
+        <version>5.16</version>
         <scope>test</scope>
       </dependency>
       <dependency>
@@ -102,6 +102,7 @@
         <artifactId>workflow-cps</artifactId>
         <version>2.32</version>
         <scope>test</scope>
+        <classifier>tests</classifier>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -168,6 +169,11 @@
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>scm-api</artifactId>
         <version>2.0.8</version>
+      </dependency>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins.workflow</groupId>
+        <artifactId>workflow-multibranch</artifactId>
+        <version>2.10</version>
         <scope>test</scope>
       </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -30,6 +30,7 @@
         <java.level>8</java.level>
         <workflow-api-plugin.version>2.28</workflow-api-plugin.version>
         <useBeta>true</useBeta>
+        <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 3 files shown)


JeremyMarshall:
So my initial objective was to codify the box in the UI to limit downstream atifact access.

So you need quite an elaborate setup to even limit that access. I'm happy to approve but I understand if the logic needs to be revisited and it never gets merged

ikedam:
Thanks for the update.
The issue sounds like https://issues.jenkins-ci.org/browse/JENKINS-24888 and more.

There’re many “does not work”s:
* No `READ` permission with `QueueItemAuthenticator`: always fail even with `CopyartifactPermission` configured.
    * I’m not sure the best way to fix this. `CopyartifactPermission` is rather for used with without `QueueItemAuthenticator`, and they don’t need to use `CopyartifactPermission` if `QueueItemAuthenticator` is properly configured.
* Specifying the upstream with a string literal: always succeed even without permissions
* Specifying the upstream with Groovy interpolation: always succeed even without permissions
* Specifying the upstream with a variable expression: always fail for the first time.


I plan this roadmap:

1. Merging this change and release copyartifact-1.41.
2. Introduce complete runtime-checks (JENKINS-24888)
3. Release copyartifact-1.42.

Good points:

* 1.42 breaks compatibility. But users can downgrade to 1.41 without any loss of configurations.
* Users can configure `copyArtifactPermission` in 1.41 and then go to 1.42. (and can back to 1.41 if something wrong).

How about this plan?

ikedam:
@JeremyMarshall @jglick 
I released copyartifact-1.41 with this change. It will be available in the update center in a day.
Thanks for the improvement!

PR #108, opened 2018-11-25
Code improvements

darxriggs:
This pull request improves the overall code quality.
I suggest to review it commit by commit as each one covers a dedicated aspect.


darxriggs:
Hi @ikedam, I updated the pull request's description.
Each commit already contains an appropriate description.

Regarding your concern on bug reports vs. code:
A bug should always be reported for a dedicated plugin version.
This version is labeled accordingly in the git repository.
Therefore I see no issue matching both.

ikedam:
Thanks for the update.

So you mean this is a part of another upcoming change for copyartifact.
Please link to that ticket to clarify that a new feature / release is planned.

jglick:
The JEP-202 work is considered finished and I do not know of any other ongoing activity in the cloud-native SIG which would affect this plugin.

darxriggs:
I have no specific changes in mind. I just had the impression that there is potential due to the current cloud native movement. After Jesse's statement, that he is finished in regard to JEP-202, the probability is much lower.

ikedam:
Though there’re useful and correct changes (e.g. fixing misspells), there’re also non-essential  changes (e.g. changing indentations, applying latest Java specifications).
Generally I prefer to preserving code histories as described before, especially without any major changes in features.

Thanks for the request, but I’ll close this request.

PR #110, opened 2019-03-25
Java 11: make sure the plugin builds fine both on JDK 8 and JDK 11

batmat:
For Java 11 readiness, apply recommended configuration:
https://wiki.jenkins.io/display/JENKINS/Java+11+Developer+Guidelines#Java11DeveloperGuidelines-MakesureyourpluginistestedinContinuousIntegrationonJava8andJava11atthesametime

And fixing what was revealed through this.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-3</version>
+    <version>1.0-beta-7</version>
   </extension>
 </extensions>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.15</version>
+        <version>3.40</version>
         <relativePath />
     </parent>
 
@@ -163,7 +163,6 @@
       <dependency>
         <groupId>org.mockito</groupId>
         <artifactId>mockito-core</artifactId>
-        <version>1.9.5</version>
         <scope>test</scope>
         <exclusions>
             <exclusion>
--- a/pom.xml
+++ b/pom.xml
@@ -28,9 +28,9 @@
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.121.1</jenkins.version>
         <java.level>8</java.level>
-        <workflow-api-plugin.version>2.28</workflow-api-plugin.version>
+        <workflow-api-plugin.version>2.33</workflow-api-plugin.version>
         <useBeta>true</useBeta>
-        <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
+        <workflow-cps-plugin.version>2.63</workflow-cps-plugin.version>
     </properties>
 
     <dependencies>
@@ -43,7 +43,7 @@
       <dependency>
           <groupId>org.jenkins-ci.plugins</groupId>
           <artifactId>structs</artifactId>
-          <version>1.14</version>
+          <version>1.17</version>
       </dependency>
       <dependency>
           <groupId>org.jenkins-ci.plugins</groupId>
@@ -120,13 +120,13 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-step-api</artifactId>
-        <version>2.13</version>
+        <version>2.19</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-support</artifactId>
-        <version>2.14</version>
+        <version>3.2</version>
         <scope>test</scope>
       </dependency>
       <dependency>
@@ -174,7 +174,7 @@
       <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>scm-api</artifactId>
-        <version>2.0.8</version>
+        <version>2.2.6</version>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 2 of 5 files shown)


batmat:
Picked up #111 so the `Jenkinsfile` here would actually be used. 
Ready for review by @jenkinsci/java11-support

batmat:
Hello @ikedam! Given you worked on this plugin recently and released it, would you please consider reviewing this?

If this works for you, once merged it would be awesome if you can also consider releasing it. Even an experimental release would be great so we can run more tests on it easily (typically https://github.com/jenkinsci/plugin-compat-tester and https://github.com/jenkinsci/acceptance-test-harness/)

Thanks!

batmat:
@ikedam gentle ping. Any way you can look at this? Thanks again!

ikedam:
@batmat Sorry, but I don’t have time to review for now, in April.
About the experimental release, I believe it’s already available when you created this request. See JEP-305 and incremental-tools for detail:
* https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc
* https://github.com/jenkinsci/incrementals-tools/blob/master/README.md#usage-in-plugin-poms

I don’t know much about Java11, and don’t know what’s the issue nor what’s the point to review (nor what’s the mandatory change).
Please let me know what’s prevented by this request. It may help me to see what’s the issue.

.mvn/extensions.xml
```diff
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-3</version>
+    <version>1.0-beta-7</version>
```
MRamonLeon: Still beta?

alecharp: yes, it's still beta..

alecharp:
@ikedam 

> I don’t know much about Java11, and don’t know what’s the issue nor what’s the point to review (nor what’s the mandatory change).
> Please let me know what’s prevented by this request. It may help me to see what’s the issue.

With the proposed `Jenkinsfile`, the plugin is tested on Java 8 and Java 11 against the `jenkins.version` and against the current LTS version of Jenkins core. To make sure it works, we had to change the parent pom version (to take the improvements we have done there) and some dependencies. 

The code change I proposed is in the test only, for the project build there is. The project was built using the Maven default JDK version setup, which wasn't working on Java 11. This because the default was Java 5 / 6. I upgraded the requirement to Java 8, as the plugin is targetting this version, and because Java 11 dropped those old versions.

pom.xml
```diff
@@ -184,6 +183,17 @@
         <scope>test</scope>
       </dependency>
     </dependencies>
+
+    <dependencyManagement>
```
oleg-nenashev: I am not sure what is the intention here. This artifact is not really consumed anywhere

alecharp: I had to add this because of some dependency upgrade which was breaking the enforcer-rule.

MRamonLeon:
Hello @ikedam, Do you have time to merge it and cut a release? Thank you.

ikedam:
Though I haven’t yet read all comments and documents carefully, here are my comments in advance about what I saw and what I can’t get yet:

* Why do you all want a new release? As @alecharp wrote, changes in this request affects only tests, and shouldn’t do nothing to released artifacts. Making a new release should result the same artifact. I suppose I mistake what “release” means.
    * These changes should not affect the plugin compatibility tester and the acceptance test harness, as they should use only the artifact. (I’m not sure, as I don’t know much about PCT or ATH).
* I tested java11 in my environment, and found the “least” changes to have tests pass with java11 is #113. I haven’t seen what other changes come from. (I’m yet not sure those differences are points to request changes).

MRamonLeon:
Hello @ikedam, thank you for your time.
We need this to have our CI environment green when running on Java 11. Whatever PR that makes it green will be suitable for us.

batmat:
@ikedam we are part of the @jenkinsci/java11-support team working on delivering Java 11 readiness to Jenkins. The initiative is described more thoroughly in https://github.com/jenkinsci/jep/blob/master/jep/211/README.adoc

Given copyartifact is used a lot (installed 46k+ times) it naturally entered a list of plugins we wanted to add Java 11 CI too.
The issue fixed with this plugin with the current release is surfacing through running copyartifact on PCT on JRE11 (see PCT codebase on https://github.com/jenkinsci/plugin-compat-tester).

So merging and releasing this PR would help Java 11 readiness for the plugin (though obviously it's not a silver bullet and won't test everything).

At least considering merging this would be great, since then it should avoid introducing new things clearly incompatible with running on a Java 11 runtime.

Thanks a lot

ikedam:
Note for me: PCT check outs source codes from the release tag: https://github.com/jenkinsci/plugin-compat-tester/blob/plugins-compat-tester-aggregator-0.0.3/plugins-compat-tester/src/main/java/org/jenkins/tools/test/PluginCompatTester.java#L544

pom.xml
ikedam: This looks to avoid:

```
[WARNING] Rule 5: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for commons-net:commons-net:3.5 paths to dependency are:
+-org.jenkins-ci.plugins:copyartifact:1.43-SNAPSHOT
  +-org.jenkins-ci.main:maven-plugin:3.1.2
    +-org.apache.maven.wagon:wagon-ftp:3.0.0
      +-commons-net:commons-net:3.5
and
+-org.jenkins-ci.plugins:copyartifact:1.43-SNAPSHOT
  +-org.jenkins-ci.main:jenkins-test-harness:2.49
    +-org.jenkins-ci.main:jenkins-test-harness-htmlunit:2.31-2
      +-commons-net:commons-net:3.6
```

and the even latest `maven-plugin` depends on `commons-net-3.5`.

Though I don’t know the best practices for avoiding `RequireUpperBoundDeps`, it seems preferable to use `exclusion` rather than `dependencyManagement` as it should work even when future jenkins-test-harness depends on `commons-net-3.7`.

batmat:
Superseded by https://github.com/jenkinsci/copyartifact-plugin/pull/113, thanks a lot @ikedam !

batmat:
> I don’t want to make a release with no changes for users

@ikedam I understand the intend, though I believe sometimes this can be useful to actually differentiate issues from one release to another. Having separate releases also helps developers, and users, to assess that a change could or couldn't have caused a regression.

In that case, releasing a minor release after the internal "Java 11 readiness" changes also allows you to see if thi causes or not issues. And then release a version with the feature, this way you can also more easily ask users to check with version X or Y to confirm an issue is present. Everybody can also save time :).

Reducing chunk size, also means reducing risk. That is one of the cool features of Continuous Delivery.

Anyway, just post I'd write this as thought, but this is your take and the bottom line is: thanks a lot for taking this over the finish line.

PR #112, opened 2019-05-03
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
+++ b/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java
@@ -394,7 +394,7 @@ public class CopyArtifact extends Builder implements SimpleBuildStep {
             // We need expand build parameters manually.
             // See JENKINS-26694, JENKINS-30357 for details.
             for(ParametersAction pa: build.getActions(ParametersAction.class)) {
-                // We have to extract parameters manally as ParametersAction#buildEnvVars
+                // We have to extract parameters manually as ParametersAction#buildEnvVars
                 // (overrides EnvironmentContributingAction#buildEnvVars)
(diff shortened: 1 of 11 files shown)


jsoref:
https://www.merriam-webster.com/dictionary/copiee
https://www.dictionary.com/misspelling?term=copiee

It is not an English word.

I spent a bit of time talking with a number of people about the best choice to replace it.

I can certainly drop the replacement if you insist. – because they're individual commits, dropping one is trivial 

As for adding words to a per repository whitelist, yes, that's more or less possible.

I'll see about adding a point about 🐪 words in the documentation, yes, that's one of the things I like about it.

ikedam:
> even though copiee is actually a correct English word

Sorry, I made a mistake in the sentence.
even though copiee is actually a “IN”correct English word

Using incorrect English words in symbol names is not critical as long as it’s intended one (pairing `copier` and `copiee` in this case), and making too many changes rather results bad such as conflict in merging other branches. And I want to have it reverted.
As I said, it’s good idea to fix them in documentaion such as Javadoc and help texts, as many users see them and they should be written in correct English.

jsoref:
@ikedam: dropped. I don't think there's anything I can usefully split out for that. If you want to pull some fraction in, the commit is 3e162a5.

jsoref:
@ikedam : I've added a section that talks about 🐫 in https://github.com/jsoref/spelling/commit/a048ca797350319c67f5b05e8555d4a4ec8c7dc7

ikedam:
Released copyartifact-1.42.1 with this fix. It will be available in the update center in a day.
Thanks for the fixes.

PR #117, opened 2020-02-29
Fix tests failing for accessing http://repo1.maven.org/

ikedam:
https://issues.jenkins-ci.org/browse/JENKINS-61276

It's caused for pom files in zip.
It's difficult to review, then

* Extract files from the zip file in the source code repository, and construct them into a zip file when running tests.
* Fix that extracted files to point to https sites.


src/test/resources/hudson/plugins/copyartifact/maven-job/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+      <id>central</id>
+      <url>https://repo.maven.apache.org/maven2</url>
+    </repository>
+  </repositories>
+  <pluginRepositories>
+    <pluginRepository>
+      <id>central</id>
+      <url>https://repo.maven.apache.org/maven2</url>
+    </pluginRepository>
+  </pluginRepositories>
```
ikedam: I added this to all pom files.
Adding this only to the root pom doesn't resolve the problem. It might caused for submodules doesn't inherit the parent module explicitly. (I don't know about detailed behavior of maven)
Adding this only to moduleA might resolve the problem as the artifacts will be downloaded for moduleA, and cached and reused for moduleB, moduleC. But that's just a workaround and not a good solution.

PR #120, opened 2020-05-23
[SECURITY-988] Fix the release date of copyartifact-1.44

ikedam:
* It was left as TBD
* Also added a link to the security advisory.


ikedam:
Just a fix for the documentations, so merge without reviewing.

PR #122, opened 2020-06-06
[JENKINS-62409] Remove matrix-auth from test dependencies

ikedam:
https://issues.jenkins-ci.org/browse/JENKINS-62409

I plan to introduce BOM.
https://github.com/jenkinsci/bom

First, remove matrix-auth from dependencies.

PR #119, opened 2020-05-11
Correct PCT failures after recent security release

Wadeck:
The failures were:

<details>

<summary>CopyArtifactPermissionPropertyTest: testNotWorkWithQueueItemAuthenticator
</summary>

```
[ERROR] testNotWorkWithQueueItemAuthenticator(hudson.plugins.copyartifact.CopyArtifactPermissionPropertyTest)  Time elapsed: 180.051 s  <<< ERROR!
org.junit.runners.model.TestTimedOutException: test timed out after 180 seconds
        at java.lang.Object.wait(Native Method)
        at java.lang.Object.wait(Object.java:502)
        at hudson.remoting.AsyncFutureImpl.get(AsyncFutureImpl.java:75)
        at org.jvnet.hudson.test.JenkinsRule.assertBuildStatus(JenkinsRule.java:1253)
        at hudson.plugins.copyartifact.CopyArtifactPermissionPropertyTest.testNotWorkWithQueueItemAuthenticator(CopyArtifactPermissionPropertyTest.java:348)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:556)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
        at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #123, opened 2020-06-13
[JENKINS-62409] Use plugin pom 4.0

ikedam:
https://issues.jenkins-ci.org/browse/JENKINS-62409

* Use plugin pom 4.0: https://github.com/jenkinsci/plugin-pom/tree/plugin-4.2
* Use jenkins bom: https://github.com/jenkinsci/bom/tree/bom-10
* Specify CI configurations explicitly instead of `recommendedConfigurations`
    * https://github.com/jenkins-infra/pipeline-library/pull/145
    * I actually don't think that's a good idea to specify configurations explicitly as we have to update Jenkinsfile manually. But Jenkins-2.164.1, the current target with `recommendedConfigurations` is old and I believe the root cause of #119 is the issue of `recommendedConfigurations`. 

I already removed the dependency to matrix-auth in #122 .

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,7 @@
 def recentLts = '2.222.4'
 
 buildPlugin(configurations: [
-  [ platform: 'linux', jdk: '8', jenkins: null ],
-  [ platform: 'windows', jdk: '8', jenkins: recentLts, javaLevel: '8' ],
-  [ platform: 'linux', jdk: '11', jenkins: recentLts, javaLevel: '8' ],
+  [ platform: 'linux', jdk: '8' ],
+  [ platform: 'windows', jdk: '8', jenkins: recentLts ],
+  [ platform: 'linux', jdk: '11', jenkins: recentLts ],
 ])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,11 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
+// Specify configurations explicitly to test against newer LTS.
+// See also https://github.com/jenkins-infra/pipeline-library/pull/145
+
+// This should be updated manually.
+def recentLts = '2.222.4'
+
+buildPlugin(configurations: [
+  [ platform: 'linux', jdk: '8', jenkins: null ],
+  [ platform: 'windows', jdk: '8', jenkins: recentLts, javaLevel: '8' ],
+  [ platform: 'linux', jdk: '11', jenkins: recentLts, javaLevel: '8' ],
+])
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.43</version>
+        <version>4.2</version>
         <relativePath />
     </parent>
 
@@ -26,14 +26,8 @@
     <properties>
         <revision>1.45</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.164.1</jenkins.version>
         <java.level>8</java.level>
-        <workflow-api-plugin.version>2.30</workflow-api-plugin.version>
-        <useBeta>true</useBeta>
-        <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
-        <git-plugin.version>3.0.5</git-plugin.version>
-        <workflow-scm-step.version>2.4</workflow-scm-step.version>
-        <workflow-multibranch.version>2.10</workflow-multibranch.version>
     </properties>
 
     <build>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -26,13 +26,13 @@
     <properties>
         <revision>1.45</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.164.3</jenkins.version>
```
ikedam, line 29: 2.164.1 is the oldest LTS available in parent-pom 4.x.
And 2.164.1 caused following test failure:
```
[ERROR] org.jvnet.hudson.test.PluginAutomaticTestBuilder$OtherTests.testPluginActive  Time elapsed: 0.037 s  <<< ERROR!
java.lang.Error: Plugin junit failed to start
        at org.jvnet.hudson.test.PluginAutomaticTestBuilder$OtherTests.testPluginActive(PluginAutomaticTestBuilder.java:99)
...
Caused by: java.io.IOException: JUnit Plugin version 1.29 failed to load.
 - You must update Jenkins from version 2.164.1 to version 2.164.3 or later to run this plugin.
        at hudson.PluginWrapper.resolvePluginDependencies(PluginWrapper.java:821)
...
```

junit-plugin 1.29 is come from maven-plugin 3.1.2, and upgrading maven-plugin can be an option.

pom.xml
```diff
@@ -26,13 +26,13 @@
...
         <java.level>8</java.level>
-        <workflow-api-plugin.version>2.30</workflow-api-plugin.version>
         <useBeta>true</useBeta>
```
ikedam, line 31: `useBeta` is still required.
```
[ERROR] hudson/plugins/copyartifact/CopyArtifact:671 jenkins/util/VirtualFile.list(Ljava/lang/String;Ljava/lang/String;Z)Ljava/util/Collection; is still in beta
[ERROR] hudson/plugins/copyartifact/CopyArtifact:689 jenkins/util/VirtualFile.readLink()Ljava/lang/String; is still in beta
[ERROR] hudson/plugins/copyartifact/CopyArtifact:696 jenkins/util/VirtualFile.toExternalURL()Ljava/net/URL; is still in beta
[ERROR] hudson/plugins/copyartifact/CopyArtifact:725 jenkins/util/VirtualFile.mode()I is still in beta
```

pom.xml
```diff
@@ -196,57 +176,36 @@
...
-        <version>${workflow-scm-step.version}</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <!-- for dependency of GitSampleRepoRule -->
-        <groupId>org.jenkins-ci.plugins.workflow</groupId>
-        <artifactId>workflow-scm-step</artifactId>
-        <version>${workflow-scm-step.version}</version>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>scm-api</artifactId>
```
ikedam, line 196: https://github.com/jenkinsci/copyartifact-plugin/pull/119/files#r439700373

ikedam:
I'll invite reviewers after I see CI passed.

ikedam:
The failure looks caused by a machine issue. Let’s retry.

src/test/java/hudson/plugins/copyartifact/ParameterizedBuildSelectorTest.java
```diff
@@ -131,7 +132,9 @@ public void testWorkflow() throws Exception {
         ));
         
         VirtualFile vf = b.getArtifactManager().root().child("artifact.txt");
-        assertEquals("foobar", IOUtils.toString(vf.open()));
+        try(InputStream in = vf.open()) {
+            assertEquals("foobar", IOUtils.toString(in, "UTF-8"));
+        }
```
ikedam, line 137: To release the file handle on Windows.
Windows denies deleting files when their handles are held.

Jenkinsfile
```diff
@@ -1 +1,11 @@
-buildPlugin(configurations: buildPlugin.recommendedConfigurations())
+// Specify configurations explicitly to test against newer LTS.
+// See also https://github.com/jenkins-infra/pipeline-library/pull/145
+
+// This should be updated manually.
+def recentLts = '2.222.4'
+
+buildPlugin(configurations: [
+  [ platform: 'linux', jdk: '8', jenkins: null ],
```
jglick, line 8:
```suggestion
  [ platform: 'linux', jdk: '8' ],
```

Jenkinsfile
```diff
@@ -1 +1,11 @@
...
+  [ platform: 'windows', jdk: '8', jenkins: recentLts, javaLevel: '8' ],
+  [ platform: 'linux', jdk: '11', jenkins: recentLts, javaLevel: '8' ],
```
jglick, line 10:
```suggestion
  [ platform: 'windows', jdk: '8', jenkins: recentLts ],
  [ platform: 'linux', jdk: '11', jenkins: recentLts ],
```

pom.xml
jglick: Yes. I do plan to remove the beta mark from at least `list` and `toExternalURL`. I am less sure about `readLink` and `mode`, which are indeed used here though my opinion is that users would be better off not relying on some filesystem-specific features.

pom.xml
```diff
@@ -26,13 +26,13 @@
...
         <java.level>8</java.level>
-        <workflow-api-plugin.version>2.30</workflow-api-plugin.version>
         <useBeta>true</useBeta>
-        <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
-        <git-plugin.version>3.0.5</git-plugin.version>
-        <workflow-scm-step.version>2.4</workflow-scm-step.version>
+        <maven-plugin.version>3.1.2</maven-plugin.version>
+        <!-- versions for test dependencies -->
+        <compress-artifacts.version>1.10</compress-artifacts.version>
+        <jenkins-test-harness-tools.version>2.0</jenkins-test-harness-tools.version>
```
jglick, line 35: Could just inline. Generally I only recommend using a POM property for a dependency version in cases where two classifiers (e.g., main and `tests`) of the same artifact are needed.

pom.xml
```diff
@@ -66,117 +66,97 @@
...
         <optional>true</optional>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>matrix-project</artifactId>
-        <version>1.7.1</version>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-support</artifactId>
```
jglick, line 92: Probably could be deleted as a transitive dep.

pom.xml
```diff
@@ -66,117 +66,97 @@
...
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-durable-task-step</artifactId>
-        <version>2.18</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-step-api</artifactId>
```
jglick, line 124: ditto

pom.xml
```diff
@@ -66,117 +66,97 @@
...
-        <version>2.13</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins.workflow</groupId>
         <artifactId>workflow-api</artifactId>
```
jglick, line 129: ditto

pom.xml
ikedam: 👍 
I definitely agree that. I wrote that in the documentation: https://github.com/jenkinsci/copyartifact-plugin#recommended-usage-of-copyartifact
But copyartifact should keep backward compatibilities and continues to use `mode` and `readLink`...

PR #125, opened 2020-07-21
[JENKINS-62267] Add test case for JENKINS-62267

basil:
See [JENKINS-62267](https://issues.jenkins-ci.org/browse/JENKINS-62267). Adds a reproducible test case for the issue, marked with `@Ignore` because the test currently fails. When the issue is resolved, the `@Ignore` annotation can be removed. I suggest looking at the fixes for [JENKINS-60393](https://issues.jenkins-ci.org/browse/JENKINS-60393) and [JENKINS-39169](https://issues.jenkins-ci.org/browse/JENKINS-39169) for inspiration as to how to fix this error.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,27 @@
         <classifier>tests</classifier>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.datadog.jenkins.plugins</groupId>
+        <artifactId>datadog</artifactId>
+        <version>1.1.2</version>
+        <scope>test</scope>
+        <exclusions>
+          <!-- to avoid RequireUpperBoundDeps with Jenkins core -->
+          <exclusion>
+            <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-constants</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>com.github.jnr</groupId>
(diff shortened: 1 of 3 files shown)


ikedam:
Thanks.

Please let me hold on merging this request as:
* It would be better to add tests without adding a dependency to datadog plugin (just as you saw, `RequireUpperBoundDeps` annoys us for every dependencies...). I suppose we can go in the same way to jenkinsci/cloudbees-folder-plugin#136.
* I'll merge the test with the fix. (So you don't need `@Ignore`)

basil:
Sounds good. Let me know if you need any additional information from me. I am also happy to test any proposed solutions.

ikedam:
I created a test code without datadog and add a fix in #126 .

I'll try this test code after merging #126.

ikedam:
Resolved conflicts and removed `@Ignore`.

ikedam:
Hmm... somehow `testRestInterfaceCannotBypassPermission` failed unexpectedly, but I could verify that `circularDependencyTest` has passed on all platforms: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fcopyartifact-plugin/detail/PR-125/2/tests

I'm ready to release 1.45.1.

basil:
Closing in favor of #126.

PR #127, opened 2020-07-25
[JENKINS-62962] Fix documentation for fingerprintArtifacts of pipeline

ikedam:
https://issues.jenkins-ci.org/browse/JENKINS-62962

* It's inconsistent with the behavior on GUI. But it cannot be fixed for backward compatibilities.


ikedam:
Documentation fix and merge without reviews.

PR #133, opened 2020-11-18
[JENKINS-61811] Make BuildSelector Serializable

rhysm:
Fixes issue when using buildSelector parameter in declarative pipeline

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/BuildSelector.java
+++ b/src/main/java/hudson/plugins/copyartifact/BuildSelector.java
@@ -33,6 +33,7 @@ import hudson.model.Job;
 import hudson.model.Run;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.io.Serializable;
 import javax.annotation.CheckForNull;
 import jenkins.util.VirtualFile;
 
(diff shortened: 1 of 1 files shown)


ikedam:
I worry that this might result an incompatibility as `BuildSelector` is an extension point and used by other plugins.

Let me know what happens if a subclass of `BuildSelector` is not serializable. (e.g. it has a field with an inserializable class)

ikedam:
link to the ticket: https://issues.jenkins.io/browse/JENKINS-61811

ikedam:
Following approach might make sense if adding `Serializable` can cause a compatibility issue:

* Mark `BuildSelectorParameter#defaultSelector`  `transient`
* Add a new field `defaultSelectorXml`, which is the XML expression of `defaultSelector`.
* Rebuild `defaultSelector` with `readResolve`.

ikedam:
@rhysm 
The issue is resolved in the way of #136 instead.
Copyartifact-1.46 with that fix will be available in the update center in a day: https://github.com/jenkinsci/copyartifact-plugin/blob/master/CHANGELOG.adoc#146
Please try that.

PR #136, opened 2020-12-19
[JENKINS-61811] Fix NotSerializableException using declarative parameter buildSelector

ikedam:
https://issues.jenkins.io/browse/JENKINS-61811

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,6 +141,12 @@
         <classifier>tests</classifier>
       </dependency>
       <dependency>
+        <groupId>org.jenkinsci.plugins</groupId>
+        <artifactId>pipeline-model-definition</artifactId>
+        <version>1.2.7</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>git</artifactId>
         <scope>test</scope>
--- a/src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
+++ b/src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
@@ -338,4 +338,34 @@ public class CopyArtifactWorkflowTest {
         );
         jenkinsRule.assertLogContains("foobar", jenkinsRule.assertBuildStatusSuccess(copier.scheduleBuild2(0)));
     }
+
+    @Issue("JENKINS-61811")
+    @Test
+    public void testBuildSelectorParameter() throws Exception {
+        WorkflowJob src = jenkinsRule.createWorkflow(
+            "TEST",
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -140,6 +140,12 @@
         <scope>test</scope>
         <classifier>tests</classifier>
       </dependency>
+      <dependency>
+        <groupId>org.jenkinsci.plugins</groupId>
+        <artifactId>pipeline-model-definition</artifactId>
+        <version>1.2.7</version>
```
ikedam, line 146: This issue happens only when using declarative pipelines.
Using the greatest version that doesn't cause `RequireUpperBoundDeps` for `workflow-multibranch`.
We can use a newer version by upgrading also the version of `workflow-multibranch`, but this version is enough to reproduce the issue.

src/test/java/hudson/plugins/copyartifact/CopyArtifactWorkflowTest.java
```diff
@@ -338,4 +338,34 @@ public void testWorkspaceBuildSelector() throws Exception {
...
+            + "    buildSelector(defaultSelector: lastSuccessful(), description: 'Build Selector', name: 'BUILD_SELECTOR')\n"
+            + "  }\n"
+            + "  stages {\n"
+            + "    stage('Test') {\n"
+            + "      steps {\n"
+            + "        copyArtifacts fingerprintArtifacts: true, projectName: 'TEST', selector: buildParameter(params.BUILD_SELECTOR)\n"
+            + "      }\n"
+            + "    }\n"
+            + " }\n"
+            + "}",
```
ikedam, line 366: A production code reported in https://issues.jenkins.io/browse/JENKINS-61811.
It's required to use `param.BUILD_SELECTOR` instead of `'BUILD_SELECTOR'`.

ikedam:
Manual migration tests (tested with Jenkins 2.263.2):

- [x] Freestyle projects with copyartifact build parameters run correctly after upgrading the plugin and restarting Jenkins (configurations are not saved yet, and migrations are performed in memory)
    * SpecificBuildSelector
    * SavedBuildSelector
    * StatusBuildSelector
- [x] Freestyle projects with copyartifact build parameters run correctly  after saving the configuration without modification with GUI
    * SpecificBuildSelector
- [x] Freestyle projects with copyartifact build parameters run correctly  after saving freestyle projects without modification with script console (`Jenkins.instance.getItemByFullName("copier", FreeStyleProject.class).save()`)
    * SavedBuildSelector
    * StatusBuildSelector
- [x] Freestyle projects with copyartifact build parameters run correctly after upgrading the configurations on the disk and restarting Jenkins.
    * SpecificBuildSelector
    * SavedBuildSelector
    * StatusBuildSelector
- [x] Declarative pipeline passes after upgrading the plugin.

PR #140, opened 2021-08-21
[JENKINS-65988] Remove usages of Guava

basil:
To avoid any potential binary incompatibilities between Guava 11 and the latest version of Guava in the forthcoming upgrade of Guava from Jenkins core, remove the Guava dependency from this plugin. This plugin doesn't need Guava; all its needs can be satisfied with native Java Platform functionality.

### DIFF
--- a/src/main/java/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor.java
+++ b/src/main/java/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor.java
@@ -125,6 +125,7 @@ public class LegacyJobConfigMigrationMonitor extends AdministrativeMonitor imple
     /**
      * @return data holding the list of jobs to warn.
      */
+    @Restricted(NoExternalUse.class)
     /* Visible for testing */ LegacyMonitorData getData() {
         return data;
     }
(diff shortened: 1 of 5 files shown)


ikedam:
References:
* https://issues.jenkins.io/browse/JENKINS-65988
* https://groups.google.com/g/jenkinsci-dev/c/aYUJ4VuOuVc/m/1JFUHJMlAQAJ

src/main/java/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor.java
```diff
@@ -126,7 +125,7 @@ public boolean isActivated() {
     /**
      * @return data holding the list of jobs to warn.
      */
-    @VisibleForTesting LegacyMonitorData getData() {
+    /* Visible for testing */ LegacyMonitorData getData() {
```
ikedam, line 128: `@Restricted(NoExternalUse.class)` is appropriate. I'll add a commit for this to this pull request later.

ikedam (approved):
Looks good to me.
Thanks a lot!
I'll push a minor change and merge after CI passed for that.

ikedam (approved):

ikedam:
CI failed for Windows EC2 instance issue. We can ignore that.

ikedam:
A new release with this fix is planned in the next weekend. Changelog: #141

ikedam:
Released in copyartifact-1.46.2.
Thanks!

PR #143, opened 2021-12-10
Compatibility with jenkins-2.325 and later (jenkinsci/jenkins#6025)

basil:
This method is changing in jenkinsci/jenkins#6025 to throw `IOException`, so callers need to be adapted.


ikedam (approved):
Looks good to me.

@basil Let me know if you know.
Does jenkins-2.325 (with  jenkinsci/jenkins#6025) fail to load copyartifact without this change?
I'm not sure we have to make a new release soon.

basil:
> @basil Let me know if you know. Does jenkins-2.325 (with [jenkinsci/jenkins#6025](https://github.com/jenkinsci/jenkins/pull/6025)) fail to load copyartifact without this change? I'm not sure we have to make a new release soon.

There is no need to make a new release soon. This PR was purely a courtesy to prevent a compilation error when bumping the core baseline in the future.

ikedam:
OK, thanks!

ikedam:

PR #144, opened 2021-12-31
Use findbugs annotations instead of jsr-305 annotations

basil:
Adapting to jenkinsci/jenkins#4604.


ikedam (approved):
Looks goot to me. Thanks.

ikedam:
`javax.annotation` was easy to remember, and `edu.umd.cs.findbugs.annotations` is so hard to remember.

ikedam:

PR #145, opened 2022-03-08
Catch both `PosixException` and `IOException`

basil:
Compatible with jenkinsci/jenkins#6345.


basil:
@ikedam Gentle ping. jenkinsci/jenkins#6345 was merged toward the forthcoming 2.339 release, so it would be nice to get this merged and released before next Tuesday in order to prevent weekly users from experiencing fatal errors when a `chmod` operation fails.

ikedam (approved):
Thanks.

I added a code comment so that we can know when to remove `PosixException`.

ikedam:

PR #147, opened 2022-04-14
Adapt icon path removal from core

NotMyFault:
The change proposed prepares the plugin for the icon path removal from core in the upcoming LTS line. This change affects plugins not using the icon API or relying on paths. Plugins using the API properly in the first place are unaffected by this change.

Could you trigger a release after merging this PR please, giving people a chance to update the plugin before updating to the next LTS version @ikedam 
Thanks in advance!

### DIFF
--- a/src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/manage.jelly
+++ b/src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/manage.jelly
@@ -95,7 +95,7 @@ THE SOFTWARE.
                                         <tr>
                                             <td class="first-column">
                                                 <j:if test="${!isValid}">
-                                                    <l:icon style="padding: 0 0 0 4px" class="icon-warning icon-sm" title="${%InvalidWarning.title}" />
+                                                    <l:icon class="icon-warning icon-sm" title="${%InvalidWarning.title}" />
                                                 </j:if>
                                             </td>
                                             <td>
--- a/src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/manage.jelly
+++ b/src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/manage.jelly
@@ -95,15 +95,15 @@ THE SOFTWARE.
                                         <tr>
                                             <td class="first-column">
                                                 <j:if test="${!isValid}">
-                                                    <img class="first-column-icon" src="${imagesURL}/16x16/warning.png" title="${%InvalidWarning.title}" />
+                                                    <l:icon style="padding: 0 0 0 4px" class="icon-warning icon-sm" title="${%InvalidWarning.title}" />
                                                 </j:if>
                                             </td>
                                             <td>
                                                 <j:if test="${jobFrom.autoMigratable}">
-                                                    <img src="${imagesURL}/16x16/accept.png" title="${%CanMigrate}" />
+                                                    <l:icon class="icon-accept icon-sm" title="${%CanMigrate}" />
                                                 </j:if>
                                                 <j:if test="${!jobFrom.autoMigratable}">
-                                                    <img src="${imagesURL}/16x16/stop.png" title="${%CanNotMigrate}" />
+                                                    <l:icon class="icon-stop icon-sm" title="${%CanNotMigrate}" />
                                                 </j:if>
                                             </td>
                                             <td>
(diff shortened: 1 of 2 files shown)


ikedam:
Thanks!
Would you link to the JIRA issue for core removing icon paths?

src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/manage.jelly
```diff
@@ -95,15 +95,15 @@ THE SOFTWARE.
                                         <tr>
                                             <td class="first-column">
                                                 <j:if test="${!isValid}">
-                                                    <img class="first-column-icon" src="${imagesURL}/16x16/warning.png" title="${%InvalidWarning.title}" />
+                                                    <l:icon style="padding: 0 0 0 4px" class="icon-warning icon-sm" title="${%InvalidWarning.title}" />
```
ikedam, line 98: Using class is preferred to using `style`.
I will see whether appropriate alternate class is available (I think I can do that this weekend).

NotMyFault:
> Would you link to the JIRA issue for core removing icon paths?

There's no JIRA issue, that already did happen months ago, I think in November last year. This is just one of the many PRs I'm proposing to maintained plugins on my list to move over to use the API.
Because a few plugins don't do that, and that's why they are affected by it.

ikedam:
OK. Thanks.

ikedam:
Looks related to this change in core: https://github.com/jenkinsci/jenkins/pull/5778

ikedam:
Since [jenkins-2.333](https://github.com/jenkinsci/jenkins/releases/tag/jenkins-2.333)

ikedam:
How to test display:

* Goto "Manage Jenkins/Configure Global Security" and switch "Copy Artifact>Compatibility mode" to "Migration"
* Create a job with copyartifact named "test"
* Goto "Manage Jenkins/Script Console", and run following command:

    ```
    hudson.plugins.copyartifact.monitor.LegacyJobConfigMigrationMonitor.get().addLegacyJob(
      Jenkins.instance.getItemByFullName("test"),
      Jenkins.instance.getItemByFullName("test"),
      new Date(),
      "username"
    );
    ```

* Access /administrativeMonitor/copyArtifactLegacyJobConfigMigration/manage
* Following icons will be displayed:
    * First column
        * If the job no longer exists (or not visible): warning icon
    * Second column
        * If the job is computed one (one with multi-branch): stop icon
        * Otherwise: accept icon

src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/manage.jelly
ikedam: .first-column-icon is the one defined here: https://github.com/jenkinsci/copyartifact-plugin/blob/copyartifact-1.46.3/src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/resources.css#L30

I'll update the css to applicable to here.

src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/resources.css
```diff
@@ -27,7 +27,8 @@
 .legacy-copy-artifact table tr td.first-column {
     padding: 0;
 }
-.legacy-copy-artifact table tr td img.first-column-icon {
+.legacy-copy-artifact table tr td.first-column .icon-warning {
+    box-sizing: content-box;
```
ikedam, line 31: * layout:icon doesn't look accept extra class. I'll work for that in https://issues.jenkins.io/browse/JENKINS-68286
* Jenkins set `box-sizing` to `border-box`: https://github.com/jenkinsci/jenkins/blob/jenkins-2.343/war/src/main/webapp/bootstrap/css/bootstrap.css#L200
    * And specifying width 16px and padding 4px results icon width 12px: https://developer.mozilla.org/ja/docs/Web/CSS/box-sizing
    * So switched to content-box here.

ikedam (approved):
Looks good to me. Thanks a lot!
I added some updates and let me know if something is wrong.
I plan to make a new release in the next weekend.

ikedam:
I verified that `grep -r -e image -e gif -e png src` hits nothing. Perfect!

src/main/resources/hudson/plugins/copyartifact/monitor/LegacyJobConfigMigrationMonitor/resources.css
ikedam: Further think, applying `padding` to icon is strange manner. I'll try `margin` instead.

ikedam:
Released with copyartifact-1.46.4

PR #150, opened 2022-08-09
Refresh plugin for 2022

basil:
### Background

See https://github.com/jenkinsci/bom/pull/1385 where we are trying to get plugin compatibility testing (PCT) up and running for this plugin.

### Problem

The plugin's build toolchain is so old that it can no longer be tested against recent Jenkins cores.

### Solution

Refresh the plugin to a modern build toolchain and Jenkins baseline. I can already hear @ikedam complaining about the baseline bump, but trends have shifted recently: https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/

### Implementation

The changes to Java code were to get newer versions of SpotBugs to pass. The changes to the `.properties` files were to pass newer versions of `InjectedTest` that enforce `.properties` files don't contain any ambiguous characters.


basil:
Used my admin powers to replay the run with the modified `Jenkinsfile` here: https://ci.jenkins.io/job/Plugins/job/copyartifact-plugin/job/PR-150/3/console

basil:
@ikedam Build is green. A quick merge and release of this isn't necessary for Jenkins users, but it would help us (core maintainers) run PCT tests in preparation for the forthcoming 2.361.1 release in September.

jglick (approved):

basil:
Hello @ikedam, as you can see in https://github.com/jenkinsci/bom/commit/564b0d8f4936d4164d47816af56b242e4be36bb9 we had to apply two workarounds to get a passing compatibility test of this plugin in `jenkinsci/bom`: testing on Java 11 rather than 17, and overriding the test harness to a newer version (1751.v11e83a_ecfe14) with a more modern HtmlUnit component that works with recent frontend changes to Jenkins weeklies. A merge and release of this PR would enable us to stop carrying these workarounds and therefore ease the burden on Jenkins core maintainers.

pom.xml
```diff
@@ -194,7 +190,7 @@
     </reporting>
 
     <scm>
-      <connection>scm:git:git://github.com/${gitHubRepo}.git</connection>
+      <connection>scm:git:https://github.com/${gitHubRepo}.git</connection>
```
ikedam, line 193: Note for me: This may affect operations during releasing.
This change is also applied to empty-plugin archetypes: https://github.com/jenkinsci/archetypes/pull/367

ikedam (approved):

ikedam:
I agree to update the target Jenkins version. I know the recommendation in the community documentation and I no longer stick to preserving target version; I did that as there were no common apparent agreements about target Jenkins version and administrators had to much costs to upgrading Jenkins version, which is almost resolved with pipelines and past security fixes.

But I have to agree that I don't like to update the target Jenkins version as it needs additional work. At least, preserving the version is easiest to maintain plugins.

> newer versions of InjectedTest that enforce .properties files don't contain any ambiguous characters.

Oops, leaving unencoded texts in comments was very useful to maintain Japanese property files...

ikedam:
@basil 
Thanks for the update.

This means that copyartifact is expected to be delivered continuously. But I won't maintain this plugin so hard.
It would be better to have you (or anyone who can continue to maintain this plugin) adopt this plugin (I believe that allows more updates, such as introducing automated releaing).
Would you consider about that? And let me know what I have to do.

I can merge this request and make a new release if the next release is expected before that. Please let me know and I think I can do that at most in two days.

jglick:
> It would be better to have you […] adopt this plugin […] Would you consider about that? And let me know what I have to do.

https://github.com/jenkins-infra/repository-permissions-updater/edit/master/permissions/plugin-copyartifact.yml and add `basil`.

pom.xml
basil: I think that changing `developerConnection` may affect operations during releasing, but not `connection` (as I am doing here).

basil:
> Oops, leaving unencoded texts in comments was very useful to maintain Japanese property files...

That will be allowed again with a baseline of 2.357 or later, and not just in comments: UTF-8 will be allowed in actual properties, too.

basil:
> This means that copyartifact is expected to be delivered continuously.

Regular updates definitely ease the burden on the core maintainers, but I would only _desire_ (and appreciate!) for plugin maintainers to do this rather than _expecting_ it. I know that there are many valid reasons for maintainers to step back (e.g., change of employment/life circumstances), and I respect that. All contributions are appreciated, no matter how small.

> Would you consider about that? And let me know what I have to do.

Sorry, but many months ago I reached my limit of plugins to maintain, so I have to decline becoming a maintainer of this plugin.

If you add the `adopt-this-plugin` label as described [here](https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/#how-can-i-mark-a-plugin-for-adoption), then anyone who is interested in maintaining this plugin can file a `repository-permissions-updater` PR without approval from you. I am aware of a team at CloudBees whose focus includes open-source plugins, and I could make that team aware of this thread if desired.

If you set up CD as described [here](https://www.jenkins.io/doc/developer/publishing/releasing-cd/), then you can simply merge PRs with a specific label and the release will be done automatically in the background.

> I can merge this request and make a new release if the next release is expected before that. Please let me know and I think I can do that at most in two days.

I wrote earlier that we found a workaround in `jenkinsci/bom`, so there is no rush to merge and release this from the perspective of `jenkinsci/bom`. While a merge and release would be appreciated from the perspective of paying down technical debt in our compatibility testing framework, there is no user-facing time pressure to deliver a new release.

ikedam:
Thanks for responses. (Sorry for my contact in weekends)
Anyway, it's better to have anyone take over maintenance of this plugin (at least, better to have anyone can do that).
I'm merging this request and making a new release, and then put this plugin "adopt-this-plugin" state.

ikedam:
Released copyartifact-1.47.

ikedam:
Oops, I don't have the permission to add a label to this repository.

basil:
@timja Could we give @ikedam the ability to add labels to this repository?

basil:
> Released copyartifact-1.47.

Thank you!

timja:
Fixed permissions

basil:
Thank you!

ikedam:
Thank you! I marked copyartifact for adoption.

PR #168, opened 2023-01-23
Use optional maven integration 3.16, not 3.19

MarkEWaite:
The plugin compatbility tester reports that the upgrade of the maven
integration plugin dependency from 3.16 to 3.19 is a problem.

https://github.com/jenkinsci/bom/pull/1703/checks is the archive of the
failing checks.

There was no strong reason to upgrade to the newer integration plugin
version in the previous release.  I upgraded because 3.19 is reasonably
recent (released 8 months ago) and after the upgrade to Jenkins 2.321+
that was done in 3.17.

Restoring the minimum dependency to 3.16 does not prevent users from
installing or using more recent releases of the maven integration plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
       <dependency>
         <groupId>org.jenkins-ci.main</groupId>
         <artifactId>maven-plugin</artifactId>
-        <version>3.19</version>
+        <version>3.16</version>
         <optional>true</optional>
       </dependency>
       <dependency>


jglick (approved):
Fine so far as it goes, but my understanding from https://ci.jenkins.io/job/Tools/job/bom/job/PR-1703/2/execution/node/3556/log/ is that the problem is simply that `maven-plugin` needs to be in the managed set, so that both this and https://github.com/jenkinsci/config-file-provider-plugin/blob/f52a38f83c6a5df5f72a25a54c803bc270182792/pom.xml#L151-L155 (and any other managed plugin with an `optional` or `test`-scoped dep) can use a BOM-managed version of `maven-plugin`.

MarkEWaite:
> the problem is simply that `maven-plugin` needs to be in the managed set

I assumed that it was desirable to not include `maven-plugin` in the managed set.  If this doesn't resolve the issue, I'm willing to attempt to add `maven-plugin` to the managed set.

jglick:
> I assumed that it was desirable to not include `maven-plugin` in the managed set.

Well, it would be great if nothing in the managed set had dependencies on it. But, so long as they do, it needs to be managed as well, or we just spend more time try to work around issues like this.

PR #203, opened 2023-08-13
Require Jenkins 2.387.3 or newer

MarkEWaite:
## Require Jenkins 2.387.3 or newer

Simplify dependency tracking in the pom file by using the maven plugin version from the plugin bill of materials instead of relying on testing of each proposed increment of the maven plugin version.  

### Testing done

Confirmed that automated tests are passing.  Since the maven plugin is now included in the bom, this change has already been checked for compatibility with bom.

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
         <revision>1.49</revision>
         <changelist>999999-SNAPSHOT</changelist>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-        <jenkins.version>2.361.4</jenkins.version>
+        <jenkins.version>2.387.3</jenkins.version>
         <useBeta>true</useBeta>
         <spotbugs.effort>Max</spotbugs.effort>
         <spotbugs.threshold>Low</spotbugs.threshold>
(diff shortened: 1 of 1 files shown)


basil:
I suppose the test failure, exposed by https://github.com/jenkinsci/jenkins-test-harness/pull/198, possibly caused the author to give up on continuing this change.

https://github.com/jenkinsci/copyartifact-plugin/pull/205 assumes this burden.

MarkEWaite:
I closed this because #205 seems like a better fix, with its improvement to tests.

basil:
And I closed #205 because it was not sufficient to deal with the fallout caused by https://github.com/jenkinsci/jenkins-test-harness/pull/198.

